PR #3, opened 2016-06-20
Add --crossgen flag.

russellhadley:
More specifically allows for use of a particular tool, and be default
is untagged (as --base and --diff's output is default tagged "base"
and "diff" respectivly.  --crossgen allows for directly placeing a
particular dasm file.  This functionality is used in the runtest script
to build up the dasm output.


russellhadley:
@pgavlin

pgavlin:
LGTM

Issue #1, opened 2016-06-16
Issues running jit-diffs on windows


AndyAyersMS:
Hit a couple of issues using this:

* Seems like JIT_DASM_ROOT must be set or app crashes.
* Windows requires .exe suffix for tools
* App should validate `git` is on path up front instead of crashing after running base set
* Framework DLL list seems outdated. Seems like you could just mine the test root.

AndyAyersMS:
@russellhadley FYI...

I may work up some fixes.

russellhadley:
@AndyAyersMS the frame work DLLs come via package so to update you only need to update the project.json, but the fx is only there to be used if CORE_ROOT isn't if you specify --core_root <blah> then that's where the framework dlls will come from.

russellhadley:
Framework update issues tracked by #13.  Closing remaining.

PR #6, opened 2016-06-27
Remove null reference exception when no JIT_DASM_ROOT specified.

russellhadley:
Change adds a bit of a better error message and fixes a initialization problem
when computing the next "tag" when the output directory is empty.

### DIFF
--- a/src/jit-diff/jit-diff.cs
+++ b/src/jit-diff/jit-diff.cs
@@ -52,6 +52,7 @@ namespace ManagedCodeGen
 
             private JObject _jObj;
             private bool _asmdiffLoaded = false;
+            private bool _noJitDasmRoot = false;
 
             public Config(string[] args)
             {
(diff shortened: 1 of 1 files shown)


russellhadley:
@AndyAyersMS @BruceForstall Fixes the issue with JIT_DASM_ROOT not specified.

russellhadley:
Fixes part 1 of issue #1

AndyAyersMS:
I think this does what I expect -- that the tool is usable via command line args even if JIT_DASM_ROOT is not set.

To confirm: is it an error if JIT_DASM_ROOT is not set, or is the app now just notifying you...?

russellhadley:
It is not an error if there is no JIT_DASM_ROOT, the message from the tool is just notification in case the user intends to have defaults via JIT_DASM_ROOT.

AndyAyersMS:
Good... changes LGTM.

PR #7, opened 2016-06-28
Update getstarted doc

JosephTremoulet:
Specify the `--frameworksonly` option that was missing in that section.

Also apply some minor formatting tweaks.


russellhadley:
LGTM

PR #8, opened 2016-07-01
Find crossgen tool in a platform independent way.

russellhadley:
This change fixes a bug on Windows where we were not finding an installed tool
due to the file extention.  We now us a globbing method to find the exe in a
platform independent way.


russellhadley:
@AndyAyersMS - Fixes .exe extension issue on windows from #1

AndyAyersMS:
LGTM. Thanks.

PR #9, opened 2016-07-01
Add try except around command creation/invocation.

russellhadley:
Adds a guard and error message around commands to make a more graceful
experience when a tool isn't on the path.

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -554,7 +554,7 @@ namespace ManagedCodeGen
             }
             catch (CommandUnknownException e)
             {
-                Console.WriteLine("\nError: git command not found!  Add git to the environment path.\n", e);
+                Console.Error.WriteLine("\nError: git command not found!  Add git to the environment path.\n", e);
                 Environment.Exit(-1);
             }
 
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -546,7 +546,18 @@ namespace ManagedCodeGen
             commandArgs.Add("--name-only");
             commandArgs.Add(diffPath);
             commandArgs.Add(basePath);
-            Command diffCmd = Command.Create(@"git", commandArgs);
+            Command diffCmd = null;
+
+            try 
+            {
+                diffCmd = Command.Create(@"git", commandArgs);
+            }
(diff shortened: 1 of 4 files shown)


russellhadley:
@AndyAyersMS - Hardens command creation and adds a more sensible error message.   Should be last element of #1.  (I'm going to open a separate issue for updating the frameworks due to reconciling the list with the .NETStandard set)

russellhadley:
@pgavlin take a look?

src/jit-analyze/jit-analyze.cs
```diff
@@ -546,7 +546,18 @@ public static bool DiffInText(string diffPath, string basePath)
...
-            Command diffCmd = Command.Create(@"git", commandArgs);
+            Command diffCmd = null;
+
+            try 
+            {
+                diffCmd = Command.Create(@"git", commandArgs);
+            }
+            catch (CommandUnknownException e)
+            {
+                Console.WriteLine("\nError: git command not found!  Add git to the environment path.\n", e);
```
pgavlin: Please use `Console.Error`

src/jit-dasm/jit-dasm.cs
```diff
@@ -515,9 +515,17 @@ public void GenerateAsm()
...
-                        commandArgs);
+                    Command generateCmd = null;
+
+                    try 
+                    {
+                        generateCmd = Command.Create(_executablePath, commandArgs);
+                    }
+                    catch (CommandUnknownException e)
+                    {
+                        Console.WriteLine("\nError: {0} command not found!\n", e);
```
pgavlin: Same

src/jit-diff/jit-diff.cs
```diff
@@ -556,6 +551,26 @@ public static int Main(string[] args)
...
+
+                // Wireup stdout/stderr so we can see outout.
+                command.ForwardStdOut();
+                command.ForwardStdErr();
+
+                return command.Execute();
+            }
+            catch (CommandUnknownException e)
+            {
+                Console.WriteLine("\nError: {0} command not found!  Add {0} to the path.", name, e);
```
pgavlin: Same

russellhadley:
@pgavlin Added separate commit for switching appropriate output to stderr.

pgavlin:
LGTM

PR #12, opened 2016-07-01
Portable apps

russellhadley:
Moves all jit utils to a flat directory of portable apps with script wrappers for invocation.

Portable apps are DLLs that have only MSIL and are invoked with 'dotnet.exe'.  Moving portable apps allows us to share the dotnet framework and only include a single directory on the path but requires wrapper scripts to hide the dotnet <mytool>.dll <args> command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,13 @@ To prepare the tools for use, from the root of the cloned jitutils repo, do the 
  2. Build and publish the tools: `build.{cmd|sh} -f -p`
 
 This will create the following directories in the repo root:
- 1. `bin` - contains one directory for each built tool.
+ 1. `bin` - contains built tools and wrapper scrips to drive them on the commandline.
  2. `fx` - contains a set of frameworks assemblies that can be used for asm diffs.
 
-Add each tool subdirectory to your path so you can easily invoke them, e.g.:
+To enable tools on the commandline just add the `bin` directory to the path.
 ```
 set jitutils=<path to root of jitutils clone>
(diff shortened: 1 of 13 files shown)


russellhadley:
@RussKeldorph

RussKeldorph:
> 

Did you consider giving this file a generic name in a central location and copying it into its final name in build.cmd/.sh?  It's not a horrible amount of duplication, but I'm not sure I see the need to have a custom one for every project, esp since they are all currently identical.

---
Refers to: src/cijobs/cijobs.cmd:3 in 2c83c61. [](commit_id = 2c83c619c60b7dd051055b4ce8717dee8dc35fec, deletion_comment = False)

src/jit-diff/jit-diff.cs
```diff
@@ -551,15 +551,24 @@ public static int Main(string[] args)
...
+        public static CommandResult TryCommand (string name, IEnumerable<string> commandArgs, bool capture = false)
         {
             try 
             {
                 Command command =  Command.Create(name, commandArgs);
 
-                // Wireup stdout/stderr so we can see outout.
-                command.ForwardStdOut();
-                command.ForwardStdErr();
+                if (capture)
```
RussKeldorph: Is this diff related to the other changes?  If not, should you at least comment on it in the description?

RussKeldorph:
:shipit:

RussKeldorph:
:+1:

PR #16, opened 2016-07-14
Add --benchmarksonly flag

JosephTremoulet:
This makes it trivial to check diffs specifically on the benchstone and
benchmark game benchmarks.


russellhadley:
Looks good.

PR #17, opened 2016-07-19
Adds relative to full path expansion for '--crossgen' flag.

russellhadley:
When the flag was addied the expansion was left out.  This change addes it in the same way as the --base or --diff flag does.


russellhadley:
@JosephTremoulet Take a look.

JosephTremoulet:
LGTM

PR #18, opened 2016-07-19
Fix path setting in bootstrap scripts

adiaaida:
The bootstrap scripts are meant to set the PATH to include the path to the
jitutils executables. However, the way the export/set commands are
written, we attempt to add the actual executable to the path, rather than
the path to the executable ($root/jitutils/bin/jit-diff instead of
$root/jitutils/bin). This change updates the command so that when the user
runs the bootstrap scripts, the correct path is added to PATH, and the
user can just run the utilities without using their full path name.

### DIFF
--- a/bootstrap.cmd
+++ b/bootstrap.cmd
@@ -22,8 +22,5 @@ popd
 
 :: set utilites in the current path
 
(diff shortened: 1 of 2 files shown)


adiaaida:
@russellhadley PTAL

russellhadley:
I think it'd be fine to remove the call to open the help page too.  It's gratuitous at this point.  Thanks for the fix!

PR #19, opened 2016-07-27
Add jit-format tool

adiaaida:


src/jit-format/jit-format.cs
```diff
@@ -0,0 +1,682 @@
...
+                else
+                {
+                    _compileCommands = Path.Combine(compileCommandsPath);
+                }
+            }
+        }
+
+        private void LoadFileConfig()
+        {
+            _jitFormatRoot = Environment.GetEnvironmentVariable("JIT_FORMAT_ROOT");
```
russellhadley: Do we want to change the format root over before initial checkin?

adiaaida: Oh right. I'll update that.

adiaaida:
/cc: @dotnet/jit-contrib

src/jit-diff/jit-diff.cs
```diff
@@ -291,11 +291,11 @@ public T ExtractDefault<T>(string name, out bool found)
...
             private void LoadFileConfig()
             {
-                _jitDasmRoot = Environment.GetEnvironmentVariable("JIT_DASM_ROOT");
+                _jitUtilsRoot = Environment.GetEnvironmentVariable("JIT_UTILS_ROOT");
 
-                if (_jitDasmRoot != null)
+                if (_jitUtilsRoot != null)
                 {
-                    string path = Path.Combine(_jitDasmRoot, "asmdiff.json");
+                    string path = Path.Combine(_jitUtilsRoot, "asmdiff.json");
```
russellhadley: Should probably change the name asmdiff.json to something like jitutils.json to cover the broader use.

src/jit-format/jit-format.cs
```diff
@@ -0,0 +1,682 @@
...
+            }
+        }
+
+        private void LoadFileConfig()
+        {
+            _jitUtilsRoot = Environment.GetEnvironmentVariable("JIT_UTILS_ROOT");
+
+            if (_jitUtilsRoot != null)
+            {
+                string path = Path.Combine(_jitUtilsRoot, "format.json");
```
russellhadley: Ahh.  I see.  I was assuming that all the format stuff would go in the json file and that there would be one file.  Is it worth combining them now?  The format stuff could be put under it's own node relatively easily if your worried about name clashes.

src/jit-format/jit-format.cs
```diff
@@ -0,0 +1,682 @@
...
+            {
+                found = true;
+
+                try
+                {
+                    return token.Value<T>();
+                }
+                catch (System.FormatException e)
+                {
+                    Console.Error.WriteLine("Bad format for default {0}.  See asmdiff.json", name, e);
```
russellhadley: Error line still notes asmdiff.json.  Should be reconciled with which ever way you go with the config file.

russellhadley:
There should be a default format.json in the root as well, or the sample asmdiff.json should be renamed/extended if the config files are unified.   Also the doc/getstarted.md document needs to be updated to reflect the format changes.  Lastly it would be nice if there was a test script that could sniff the jit-format exe.  So anyone that extends it can be minimally sure that the base functionality is still there.  There's a limited set of examples there now for jit-dasm/jit-analyze.

src/jit-format/jit-format.cs
adiaaida: What sorts of additional changes would need to be made to the json parsing to make that work? I don't really know much about this, so I just went with duplicating the logic with its own file, since I knew I could make that work.

russellhadley: The jObj ends up looking much like a map or associative array.  You could go 2 ways.  You could add a 'format' node with either 'default' as a sub node or with the key/value pairs you want directly under it.  Which would also mean that jit-diff should be changed to have a 'asmdiff' node with a sub 'default' node.  Or you could combine all the defaults into a single node.  I think I'm inclined to the first rather than the second since I think it'd be cleaner.

src/jit-format/jit-format.cs
```diff
@@ -0,0 +1,682 @@
...
+                    string[] splitLine = line.Trim().Split(new string[]{ "CL.exe " }, StringSplitOptions.None);
+                    clExe = (splitLine[0] + "CL.exe").Replace("\\", "/");
+
+                    string[] options = splitLine[1].Split(' ');
+
+                    bool dOptionFound = false;
+
+                    foreach (string option in options)
+                    {
+                        if (option.StartsWith("/I"))
```
RussKeldorph: >/ [](start = 47, length = 1)

cl.exe accepts options beginning with "-" as well as "/" so you should check for both even if it happens that we don't see it today.

src/jit-format/jit-format.cs
```diff
@@ -0,0 +1,682 @@
...
+                compileCommandsWriter.WriteLine("]");
+                compileCommandsWriter.WriteLine();
+            }
+        }
+
+        public static bool RunClangTidy(List<string> filenames, string compileCommands, bool fix, bool verbose)
+        {
+            bool formatOk = true;
+            string checks = "readability-braces*,modernize-use-nullptr";
+            string tidyFix = fix ? "-fix " : " ";
```
RussKeldorph: I feel like this ought to be part of the config file rather than baked into the code.  Seems like something it will be nice to change without rebuilding this tool.  Perhaps not urgent for v1.0, but something to do eventually.

src/jit-format/jit-format.cs
```diff
@@ -0,0 +1,682 @@
...
+        public bool BuildCompileCommands { get { return _buildCompileCommands; } }
+        public string CoreCLRRoot { get { return _rootPath; } }
+        public string BuildLog { get { return _buildLog; } }
+        public string CompileCommands { get { return _compileCommands; } }
+        public IReadOnlyList<string> Filenames { get { return _filenames; } }
+        public IReadOnlyList<string> Projects { get { return _projects.Count == 0 ? new List<string>{"dll"} : _projects; } }
+        public string SourceDirectory { get { return _srcDirectory; } }
+    }
+
+    public class jitformat
```
RussKeldorph: >jitformat [](start = 17, length = 9)

Is this casing intentional?  You would normally use Pascal casing for public identifiers.

src/jit-format/jit-format.cs
adiaaida: Got it. Will fix.

src/jit-format/jit-format.cs
adiaaida: I could add it as an optional option with these as the default if none are provided? That's an easy change.

src/jit-format/jit-format.cs
adiaaida: I followed the example of the other tools, so yes it was intentional.

src/jit-format/jit-format.cs
RussKeldorph: Probably best to leave for a later change once we've decided what it will look like.

src/jit-format/jit-format.cs
RussKeldorph: Ok, consistency is most important.  TBD whether this casing is right.  I suppose @russellhadley thinks so. :)

russellhadley: For the main class, though odd, it was done to match the commandline on linux (case sensitive).  I thought it a special case.

adiaaida:
This is the current version that I've run locally for windows x64. Through some testing, we found that clang-tidy has errors trying to parse x86 because it uses ms style asm that clang doesn't support.

This also needs some updates for linux (we need to modify the compile_commands database), and for dealing with incremental builds.

bootstrap.cmd
```diff
@@ -3,6 +3,7 @@
 where /q dotnet.exe || echo Can't find dotnet.exe! Please add to PATH && goto :EOF
 where /q git.exe || echo Can't find git.exe! Please add to PATH && goto :EOF
 set root=%~dp0
+pushd %root%
```
russellhadley: I can't seem to think of why we need this, what am I missing?

russellhadley:
@adiaaida, thanks for adding the test and updating the getstarted.md doc to cover the changes for the JIT_UTIL_ROOT.  I'm still wondering where we will put the directions for running the formatting tool. It might be easiest to just add it as a section to the getstarted.md and then point to it from a doc in coreclr/doc.  @RussKeldorph what do you think?

bootstrap.cmd
adiaaida: It was for how I was planning on running this in the lab. I don't know if it's necessarily correct, but I wanted to be able to call bootstrap as a peer of coreclr (making jitutils a peer), but call it without having to change directories in the lab job. I just didn't like the idea of having jitutils as a subdirectory of coreclr in the lab job.

RussKeldorph:
@russellhadley Yes, leaving the primary docs here and linking to them sounds good.  The tool itself should probably refer there as well in its usage output.

bootstrap.cmd
RussKeldorph: If you have to change directories to run some other script, that almost always indicates a bug in the other script.  I'd like to understand more about this requirement.

russellhadley: Seems fine.  The semantics are "put the jitutils dir where the script is" right?  Rather than where it's invoked from.

adiaaida: Yes, exactly.

adiaaida:
I added additional documentation to getstarted.md.

PTAL

doc/getstarted.md
```diff
@@ -450,3 +471,117 @@ usage: jit-diff install [-j <arg>] [-n <arg>] [-l] [-b <arg>]
...
+                                protojit, etc.). Default: dll
+    <filenames>...              Optional list of files that should be
+                                formatted.
+```
+
+A common task for developers would be to format their changes before submitting a
+PR:
+
+```
+jit-format -a x64 -b Debug -o Windows_NT -c C:\michelm\coreclr
```
RussKeldorph: This needs to run after a build, right?  Is that mentioned anywhere?

RussKeldorph:
LGTM.

doc/getstarted.md
adiaaida: Right. I'll update.

adiaaida:
@russellhadley ?

adiaaida:
@russellhadley PTAL

src/jit-format/jit-format.xproj
```diff
@@ -0,0 +1,19 @@
...
+    <RootNamespace>mcgdiff</RootNamespace>
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">.\obj</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">.\bin\</OutputPath>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <SchemaVersion>2.0</SchemaVersion>
+  </PropertyGroup>
+  <Import Project="$(VSToolsPath)\DotNet\Microsoft.DotNet.targets" Condition="'$(VSToolsPath)' != ''" />
+</Project>
```
russellhadley: Nit.  Line ending so git doesn't nag about it.

russellhadley:
Added one more nit, but LGTM.  We should figure out where to update/post the bootstrap directions.  Particularly since the semantics changed with this checkin.

PR #21, opened 2016-08-02
Update cijobs to allow a different repo and different product path.

ianhays:
This commit modifies two of the hardcoded values so that cijobs can be used in CoreFX.

The values changed are the path to the zip to pull down, and the repo (product) name. Both are now passable as additional parameters, though the defaults are set to what they used to be so that the new program will be backwards-compatible with all existing uses.

@russellhadley PTAL

 @ramarag the binaries for this are available at `%_ARM64Root%\common\tools\cijobs\bin\Debug\netcoreapp1.0\win10-arm64\publish\`. I've copied over the necessar win10-arm64 bits, though you'll still need to run the dll on top of corerun as the exe in that folder is win7. Example:
```%_ARM64Root%\common\tools\cijobs\bin\Debug\netcoreapp1.0\win10-arm64\publish\corerun %_ARM64Root%\common\tools\cijobs\bin\Debug\netcoreapp1.0\win10-arm64\publish\cijobs.dll copy --unzip --job arm64_cross_release_windows_nt --number 1187 --output d:\scratch --ContentPath <depends on what you're archiving in your netci file> --repo dotnet_coreclr```.

That command would download whatever binaries you decide to archive for [this job](http://dotnet-ci.cloudapp.net/job/dotnet_coreclr/job/master/job/arm64_cross_release_windows_nt/1187/) and unzip them into `d:\scratch'. Feel free to stop by if you've got any questions about using it in CoreCLR.

### CODE REVIEW

File: src/cijobs/project.json [hunk 1]
```diff
@@ -15,11 +15,13 @@
...
         "dependencies": {
             "Microsoft.NETCore.App": {
-                "type": "platform",
                 "version": "1.0.0"
             }
         },
         "imports": "dnxcore50"
     }
+  },
+  "runtimes": {
```
russellhadley: I don't think we want to check in making this a standalone configuration.  We should probably figure out a way to build them both ways.  (ideas?)

russellhadley: Note that cijobs has to run all all targets (mac, flavors of linux, etc. so turning this back into a standalone means you need to add lots more to the runtime section)

ianhays: Yeah I was wondering about that. Technically it doesn't really matter what's in the repo since we have to build, replace binaries for win10-arm64, and copy to the arm64 share manually regardless. 

I'll revert the project.json change. Going from shared->standalone is a small enough change that I think the drift is acceptable in this case.

src/cijobs/cijobs.cs
```diff
@@ -523,22 +534,22 @@ private class CopyCommand
...
                     // unzip archive in place.
-                    var zipPath = Path.Combine(outputPath, "Product.zip");
+                    var zipPath = Path.Combine(outputPath, Path.GetFileName(contentPath));
                     ZipFile.ExtractToDirectory(zipPath, outputPath);
                 }
             }
         }
     }
-}
+}
```
russellhadley: This is a nit, but could you add CR here so there isn't the little git nag about it being missing?

ianhays: sure

russellhadley:
Looks good :)

PR #22, opened 2016-08-05
Add approximate overload counts to the "top N" methods lists

AndyAyersMS:
Jit diffs are unable to disambiguate some kinds of overloads because types
are mapped into a simplified set (ref, struct, etc). This can make
interpreting the total bytes of diff per method a bit tricky since the
entry may represent the total change from a set of methods. So, report on
how many folded overloads there are in the BASE set.

Example:
```
Top method improvements by size (bytes):
       -2801 : System.Private.CoreLib.dasm - System.Array:LastIndexOf(ref,struct,int,int):int (9 methods)
       -2328 : System.Private.CoreLib.dasm - System.Array:Sort(ref,int,int,ref) (25 methods)
```

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -393,7 +393,8 @@ namespace ManagedCodeGen
                                         {
                                             path = fd.path,
                                             name = md.name,
-                                            deltaBytes = md.deltaBytes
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@russellhadley PTAL
cc @dotnet/jit-contrib

erozenfeld:
LGTM

PR #23, opened 2016-08-11
Add downloads of clang-format and clang-tidy; Add SOURCE_FORMATTING

adiaaida:
This change does the following:

1) Updates bootstrap.cmd to download the windows versions of clang-format and clang-tidy. This is for lab support (and also for devs who don't want to install the tools manually).

2) Add -DSOURCE_FORMATTING to the compile_commands.json to correspond with the #ifdef that was added to include files to allow clang-tidy to parse jit sources.

3) Add a link to the getstarted.md doc if clang-tidy or clang-format fail.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -26,6 +26,36 @@ dotnet restore
 
 ./build.sh -p -f
 
+if ! which -s clang-format || ! which -s clang-tidy;
+then
+
+    info=$(dotnet --info)
+
+    if echo $info | grep -q -i 'osx';
+    then
+        # download osx version of clang-tidy/format
+        wget https://clrjit.blob.core.windows.net/clang-tools/osx/clang-format.exe -O bin/clang-format.exe
+        wget https://clrjit.blob.core.windows.net/clang-tools/osx/clang-tidy.exe -O bin/clang-tidy.exe
+    elif echo $info | grep -q -i 'ubuntu.16.04';
+    then
+        # download osx version of clang-tidy/format
+        wget https://clrjit.blob.core.windows.net/clang-tools/ubuntu/16.04/clang-format.exe -O bin/clang-format.exe
+        wget https://clrjit.blob.core.windows.net/clang-tools/ubuntu/16.04/clang-tidy.exe -O bin/clang-tidy.exe
+    elif echo $info | grep -q -i 'ubuntu';
+    then
+        # download osx version of clang-tidy/format
+        wget https://clrjit.blob.core.windows.net/clang-tools/ubuntu/14.04/clang-format.exe -O bin/clang-format.exe
+        wget https://clrjit.blob.core.windows.net/clang-tools/ubuntu/14.04/clang-tidy.exe -O bin/clang-tidy.exe
+    elif echo $info | grep -q -i 'centos';
+    then
+        # download osx version of clang-tidy/format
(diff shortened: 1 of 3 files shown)


adiaaida:
@RussKeldorph @russellhadley PTAL

src/jit-format/jit-format.cs
```diff
@@ -580,52 +589,61 @@ public static bool RunClangTidy(List<string> filenames, string compileCommands,
...
-                    string stdoutx = process.StandardOutput.ReadToEnd();         
-                    string stderrx = process.StandardError.ReadToEnd();             
-                    process.WaitForExit();
-
-
-                    if ((stdoutx.Contains("warning:") || (!ignoreErrors && stdoutx.Contains("error:"))))
+                    try
                     {
-                        if (verbose)
+                        Process process = new Process();
```
russellhadley: Must have missed this some how the first time.  Why didn't you use the simpler command invocation wrapper that was used in the other tools?

bootstrap.sh
```diff
@@ -26,6 +26,34 @@ dotnet restore
...
+    wget https://clrjit.blob.core.windows.net/clang-tools/ubuntu/14.04/clang-tidy.exe -O bin/clang-tidy.exe
+elif echo $info | grep -q -i 'centos';
+then
+    # download osx version of clang-tidy/format
+    wget https://clrjit.blob.core.windows.net/clang-tools/centos/clang-format.exe -O bin/clang-format.exe
+    wget https://clrjit.blob.core.windows.net/clang-tools/centos/clang-tidy.exe -O bin/clang-tidy.exe
+else
+    echo "Clang-tidy and clang-format not installed. Please install and put them on the PATH to use jit-format."
+fi
+
```
russellhadley: So it's not just the windows bootstrap that downloads the tools.   Should there be some kind of check to make sure that they're not already on the path?

src/jit-format/jit-format.cs
adiaaida: This is new. You didn't miss it. Sorry.

Let me look into this simpler command invocation wrapper.

bootstrap.sh
adiaaida: Can do.

RussKeldorph: I believe we are relying on a specific version of the tools, so we should do a version check if we don't unconditionally just use these.  In any case, it probably makes sense to output a message that we are downloading these.

bootstrap.sh
```diff
@@ -26,6 +26,34 @@ dotnet restore
...
+    # download osx version of clang-tidy/format
+    wget https://clrjit.blob.core.windows.net/clang-tools/ubuntu/14.04/clang-format.exe -O bin/clang-format.exe
+    wget https://clrjit.blob.core.windows.net/clang-tools/ubuntu/14.04/clang-tidy.exe -O bin/clang-tidy.exe
+elif echo $info | grep -q -i 'centos';
+then
+    # download osx version of clang-tidy/format
+    wget https://clrjit.blob.core.windows.net/clang-tools/centos/clang-format.exe -O bin/clang-format.exe
+    wget https://clrjit.blob.core.windows.net/clang-tools/centos/clang-tidy.exe -O bin/clang-tidy.exe
+else
+    echo "Clang-tidy and clang-format not installed. Please install and put them on the PATH to use jit-format."
```
RussKeldorph: This message should recommend a specific version and/or URL.

adiaaida:
Closing this one because a lot of these changes have been subsumed by other PRs.

PR #25, opened 2016-08-15
Fix compile_commands.json for non-Windows

adiaaida:
On non-Windows platforms, we use CMAKE_EXPORT_COMPILE_COMMANDS, which
creates commands for every file and every project. We are only
interested in jit files, and we need each project to be separated so
that clang-tidy does not attempt to run on each file multiple times. If
it does, it replaces NULL with nullptr several times and we end up with
nullptrptrptr all over the code. This change takes the compile_commands
database generated by CMake, parses it and writes a new compile_commands
database with only the commands we are currently interested in.


adiaaida:
@pgavlin @russellhadley @RussKeldorph PTAL
/cc: @dotnet/jit-contrib

src/jit-format/jit-format.cs
```diff
@@ -563,6 +568,54 @@ public static void ParseBuildLog(string buildLog, string compileCommands, string
             }
         }
 
+        public static string rewriteCompileCommands (string compileCommandFile, string project)
+        {
+            string[] compileCommands = File.ReadAllLines(compileCommandFile);
```
RussKeldorph: I can see that only relatively primitive parsing is required at this point, but why not parse as JSON?

src/jit-format/jit-format.cs
```diff
@@ -563,6 +568,54 @@ public static void ParseBuildLog(string buildLog, string compileCommands, string
             }
         }
 
+        public static string rewriteCompileCommands (string compileCommandFile, string project)
```
RussKeldorph: Please add a comment header describing the transformation this method intends to make.

src/jit-format/jit-format.cs
adiaaida: I was having a hard time finding documentation to help me figure out the errors I was seeing when I tried to parse it as JSON, so I went with this.

adiaaida: I think I figured out how to parse the JSON. I'll update this in a few minutes

adiaaida:
Updated the parser to use a Json parser.

PTAL.

src/jit-format/jit-format.cs
```diff
@@ -563,6 +582,36 @@ public static void ParseBuildLog(string buildLog, string compileCommands, string
...
+                    newCommands.Add(new CompileCommand(command["directory"].Value<string>(),
+                                command["command"].Value<string>(), command["file"].Value<string>()));
+                }
+
+            }
+
+            // write commands back to a file.
+            string newCompileCommandsFileName = Path.Combine(Path.GetDirectoryName(compileCommandFile), "compile_commands_" + project + ".json");
+            string json = JsonConvert.SerializeObject(newCommands.ToArray(), Formatting.Indented);
+            System.IO.File.WriteAllText(newCompileCommandsFileName, json);
```
pgavlin: Thank you for not overwriting the existing compile commands file :)

adiaaida: You betcha.

src/jit-format/jit-format.cs
```diff
@@ -343,6 +343,20 @@ public T ExtractDefault<T>(string name, out bool found)
         public string SourceDirectory { get { return _srcDirectory; } }
     }
 
+    public class CompileCommand
```
RussKeldorph: Does this need to be public?

src/jit-format/jit-format.cs
```diff
@@ -644,11 +693,11 @@ public static bool RunClangFormat(List<string> filenames, bool fix, bool verbose
 
                 if (verbose)
                 {
-                    Console.WriteLine("Running: clang-format {0} {1}", formatFix, filename);
+                    Console.WriteLine("Running: clang-format {0} -style=file {1}", formatFix, filename);
```
RussKeldorph: Is this fix related?  If not, it should either be separated into its own change or accounted for in the description.

src/jit-format/jit-format.cs
```diff
@@ -563,6 +582,36 @@ public static void ParseBuildLog(string buildLog, string compileCommands, string
...
+
+            foreach (JObject command in commands.Children<JObject>())
+            {
+                // Search for directory entries containing jit/<project>
+                if (command["directory"].Value<string>().Contains("jit/" + project))
+                {
+                    // Add the command to our list of new commands
+                    newCommands.Add(new CompileCommand(command["directory"].Value<string>(),
+                                command["command"].Value<string>().Replace("-I", "-isystem"), command["file"].Value<string>()));
+                }
```
RussKeldorph: Use local temps to break up line and make it easier to read.

RussKeldorph:
:shipit:

src/jit-format/jit-format.cs
adiaaida: No, it isn't. I'm not 100% sure it's necessary, though @pgavlin pointed it out on Friday. I've tried clang-format with and without -style=file, and I get the same results, so it's just a bit of paranoia

adiaaida:
Apparently, clang-tidy doesn't actually like it when you give it a compile commands database that isn't called compile_commands.json. So I am going to have to change up the logic a little bit.

adiaaida:
I have updated this to do the following to workaround the strange clang-tidy behavior I was seeing:

1. Move the compile_commands.json to compile_commands_full.json.
2. Parse compile_commands_full, and save away the results to compile_commands.json
3. Run clang-tidy.
4. Delete compile_commands.json and then move compile_commands_full back to compile_commands.json.

adiaaida:
PTAL at the new changes.

src/jit-format/jit-format.cs
RussKeldorph: Ok, please update the description.

RussKeldorph:
LGTM

PR #26, opened 2016-08-17
Use command

adiaaida:
This change includes the following, split into 3 commits: 

1. Change the non-fix mode of DoClangTidy to run in parallel.

2. Use the builtin Command API for clang-tidy improves the readability of the code, and makes jit-format fail gracefully when clang-tidy does not exist

3. Command strips out blank lines when capturing standard output from processes it runs, so we need a workaround for clang-format in order to use this functionality. Clang-format has an -output-replacement-xml that we can use which prints the fixes in an xml format. We can parse this xml, and use it to determine if there are formatting errors. We can also use it to calculate and print out the changes that would need to be made, making this workaround a more verbose method. This change also changes DoClangFormat to run in parallel.

The new output will look like this (comments are mine):

clang-format: there are formatting errors in /Users/adiaaida/code/coreclr/src/jit/alloc.cpp
```
At Line 106 Before:
{
      assert(getDefaultPageSize()  !=  0); // 6 spaces of indentation
After:
{
    assert(getDefaultPageSize()  !=  0); // indentation is fixed to 4 spaces

At Line 107 Before:
    assert(getDefaultPageSize()  !=  0); // extra space before the !=
After:
    assert(getDefaultPageSize() !=  0); // extra space removed

At Line 107 Before:
    assert(getDefaultPageSize() !=  0); // extra space after the != (also note that the fix above has already been applied here)
After:
    assert(getDefaultPageSize() != 0); // extra space removed

At Line 107 Before:
    assert(getDefaultPageSize() != 0); // note this line has now had both fixes applied.
  assert(isInitialized()); // 2 spaces of indentation
After:
    assert(getDefaultPageSize() != 0);
    assert(isInitialized()); // indentation fixed to 4 spaces
```


adiaaida:
@RussKeldorph PTAL
/cc: @dotnet/jit-contrib

src/jit-format/jit-format.cs
```diff
@@ -640,66 +642,122 @@ public static bool RunClangTidy(List<string> filenames, string compileCommands,
...
+                                Console.WriteLine(stdoutx);
+                            }
                         }
                     }
                 }
             }
+            else
+            {
+                Parallel.ForEach(filenames, (filename) =>
+                        {
```
RussKeldorph: It's tough to tell just looking at this diff, but there appears to be a substantial amount of duplication between this loop and the one above that needs to be factored into helper methods.  You should either control the differences in behavior with method parameters or move as much as you can into one or more methods and leave the differences inside the loop bodies.  If that's not clear enough, let me know and I'll take a closer look.

RussKeldorph: Oh, and I think this block is indented too far.

adiaaida: There is quite a bit of duplication. I'll factor it out.

adiaaida: Done. PTAL.

RussKeldorph:
@adiaaida It looks like your refactoring modified the last commit instead of the "Parallelize foreach in DoClangTidy" commit.  Not the end of the world, but in the future going through the Git gynastics required to fix the history is worth it.

src/jit-format/jit-format.cs
```diff
@@ -638,130 +639,169 @@ public static bool RunClangTidy(List<string> filenames, string compileCommands,
...
+        }
+
+        public static bool RunClangFormat(List<string> filenames, bool fix, bool verbose)
+        {
+            string formatFix = fix ? "-i" : "";
+            string outputReplacementXml = fix ? "" : "-output-replacements-xml";
+            bool formatOk = true;
+
+            Parallel.ForEach(filenames, (filename) =>
                         {
```
RussKeldorph: Why do these get indented so far?

adiaaida: It was a mistake. Sorry.

adiaaida:
I intended the change to modify the "Use Command API instead of Process for DoClangTidy" commit. I think that's the one that did get updated. Am I wrong?

RussKeldorph:
@adiaaida Well, regardless, I made the comment in the "Parallelize foreach in DoClangTidy" commit, so that's the one I would have expected to be modified. :)

adiaaida:
@RussKeldorph I see. I did it where I did it since the next commit updates the logic used in the for loops. Do you want me to mess around with it?

RussKeldorph:
@adiaaida No, I will figure it out.

src/jit-format/jit-format.cs
```diff
@@ -638,66 +639,67 @@ public static bool RunClangTidy(List<string> filenames, string compileCommands,
...
-                    }
-                        
+        public static bool DoClangTidyInnerLoop(bool fix, bool ignoreErrors, string checks, string compileCommands, string filename, bool verbose)
+        {
+            string tidyFix = fix ? "-fix" : "";
+            string fixErrors = ignoreErrors && fix ? "-fix-errors" : "";
+
+            bool formatOk = true;
+
+            if (filename.EndsWith(".cpp"))
```
RussKeldorph: Random note not related to this change: It's a bit unfortunate to hardcode this.  You don't need to change it since it's probably good enough for now, but it would be nice if this were derived from the compile_commands.json since that naturally filters out files that aren't directly compiled like headers.

src/jit-format/jit-format.cs
```diff
@@ -706,62 +708,100 @@ public static bool RunClangTidy(List<string> filenames, string compileCommands,
...
+                            // Read in the xml generated by clang-format
+                            var doc = new XmlDocument();
+                            doc.LoadXml(result.StdOut);
+
+                            XmlNodeList replacements = doc.DocumentElement.ChildNodes;
+
+                            Console.WriteLine("");
+                            Console.WriteLine("clang-format: there are formatting errors in {0}", filename);
+
+                            foreach (XmlNode replacement in replacements)
```
RussKeldorph: Would you change this back to the simpler version if the CLI bug that swallows newlines were fixed?  If you would, then you should add a comment that this relatively large block of code is here because we are working around a bug.

adiaaida: I think probably not. The nice thing about this change is that we can actually show the user all of the formatting errors that they have in each file, not just the first formatting error.

RussKeldorph:
This was a bit hard to review because IMO the first and second commits got tangled up, so I don't have great confidence I didn't miss something, but aside from some minor comments LGTM.

src/jit-format/jit-format.cs
RussKeldorph: Ok.

src/jit-format/jit-format.cs
adiaaida: This is more of a paranoia check for the cases where the user specifies which files to run on. In that case, we'll still need to make sure that we're only running on cpp files.

PR #27, opened 2016-08-17
Download clang-format/clang-tidy if not on PATH

adiaaida:
This change updates boostrap.cmd/sh to install clang-format/clang-tidy
if they are not currently on the PATH. We then check the version of the
tools that are installed to confirm that version 3.8.* is installed. If
the wrong version is installed, boostrap.cmd/sh will inform the user
that jit-format requires 3.8.*, and will direct them to LLVM.org to
install the appropriate version.

We currently have downloads of the tools on Azure blob storage for
Ubuntu 14.04, Ubuntu 16.04, OSX, CentOS, and Windows. On other operating
systems, we direct the user to LLVM.org to install the appropriate
version for their platform.

### DIFF
--- a/bootstrap.cmd
+++ b/bootstrap.cmd
@@ -19,6 +19,34 @@ dotnet restore
 
 call .\build.cmd -p -f
 
+where /Q clang-format
+IF %errorlevel% NEQ 0 GOTO DownloadTools
+
+where /Q clang-tidy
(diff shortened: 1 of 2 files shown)


adiaaida:
/cc: @dotnet/jit-contrib

russellhadley:
LGTM

RussKeldorph:
LGTM

PR #28, opened 2016-08-18
Fix SetPlatform() and bootstrap for Linux platforms

adiaaida:
I assumed that coreclr built into the name of the operating system on
various flavors of Linux, but it turns out they are built into
Linux.<arch>.<build>. This change fixes SetPlatform to set the OS as
Linux if the OS reported by dotnet --info is not Windows or OSX.

Additionally, in bootstrap, we were trying to get the exes which do
not exist for Linux platforms (they do not have extensions). This
change fixes the wget, and makes the files executable. It also only
checks the version if the tools were already on the path.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,5 +1,14 @@
 #Quick and dirty bootstrap. 
 
+function validate_url(){
+  if [[ `wget -S --spider $1  2>&1 | grep 'HTTP/1.1 200 OK'` ]];
+  then
+      return 0;
+  else
+      return 1;
+  fi
+}
+
 if ! dotnet --info; then 
     echo "Can't find dotnet! Please add to PATH."
     return 1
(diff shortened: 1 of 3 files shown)


adiaaida:

bootstrap.sh
```diff
@@ -35,39 +35,47 @@ then
     then
         echo "Downloading clang-tidy and clang-format to bin directory"
```
RussKeldorph: Do we control the directory structure of clang-tools?  If so, why not make the folders match the RID so you just use it to directly derive the path name.  Fail with a useful message if the dir doesn't exist.  Then, to add new platform support,  we just upload some files to the blob and the script doesn't need to change.

If we can't do that, we should refactor this code to be data-driven.  Create a table of pairs ("osx" -> "osx", "ubuntu.16.04" -> "ubuntu/16.04"...) that will let you derive the name.

At the very least, the common code in these blocks should be pulled out into a helper that takes platform ID and directory fragment as parameters.

bootstrap.sh
```diff
@@ -35,39 +35,47 @@ then
...
-then
-    echo "jit-format requires clang-format and clang-tidy version 3.8.*. Currently installed: "
-    clang-format --version
-    clang-tidy --version
-    echo "Please install version 3.8.* and put the tools on the PATH to use jit-format."
-    echo "Tools can be found at http://llvm.org/releases/download.html#3.8.0"
+else
+    if ! clang-format --version | grep -q 3.8;
+    then
+        echo "jit-format requires clang-format and clang-tidy version 3.8.*. Currently installed: "
```
RussKeldorph: Have you confirmed there are no formatting differences between 3.8.0 and 3.8.1?  We shouldn't allow wildcards if there are any differences.

adiaaida: At least on Windows, we didn't notice any differences when Pat used 3.8.1 compared to my 3.8.0.

bootstrap.sh
adiaaida: We do control the directory structure. I can change this to do that, at least in bootstrap.sh pretty easily, I think.

bootstrap.sh
RussKeldorph: If we're currently formatted with 3.8.1, then I think I would lock everybody into that specific version with this check.  Doesn't seem to cost much, and it could have a benefit.

src/jit-format/jit-format.cs
```diff
@@ -104,7 +104,8 @@ private void SetPlatform()
                         }
                         else
                         {
-                            _os = match.Groups[1].Value.Trim();
+                            // Assuming anything other than Windows or OSX is a Linux flavor
+                            _os = "Linux";
```
RussKeldorph: Why not match "Linux" or whatever for this case and then error gracefully in another else clause?  We already know xxxBSDs are out there.

adiaaida: We're matching against OS Name, which varies based on OS. If I looked at OS platform, I could do that.

bootstrap.sh
adiaaida: Does anyone have a CentOS machine on the team?

src/jit-format/jit-format.cs
RussKeldorph: That sounds fine.  My main gripe is a case being used as a catch-all when we know it's not.  We should match what we know we want to support and gracefully error out on other cases, even if there's a chance the catch-all would actually do the right thing.

bootstrap.sh
RussKeldorph: I have a VM.  What do you want to know?

adiaaida: What is the RID

bootstrap.sh
adiaaida: 3.8.1 is only available for certain platforms. 3.8.0 is available for all, so we need to lock into that version for now.

RussKeldorph: Ok.

bootstrap.sh
RussKeldorph: Takes a while to boot.  Will get back to you...

RussKeldorph: Output of dotnet --info:
.NET Command Line Tools (1.0.0-preview2-003121)

Product Information:
 Version:            1.0.0-preview2-003121
 Commit SHA-1 hash:  1e9d529bc5

Runtime Environment:
 OS Name:     centos
 OS Version:  7
 OS Platform: Linux
 RID:         centos.7-x64


---
In reply to: [75393080](https://github.com/dotnet/jitutils/pull/28#discussion_r75393080) [](ancestors = 75393080)

adiaaida: Thank you!

adiaaida:
Fixed comments.

PTAL.

RussKeldorph:
LGTM

PR #29, opened 2016-08-19
Add more robust platform detection and checking of defaults

russellhadley:
Fixes some usability bugs on Linux that showed up in daily use.

- Fix format string in cijobs that cropped up from Ian's change.
- Fix issue with installing tools on Linux where we weren't defaulting to the right string to find the product tools folder downloaded from the coreclr build.
- Check paths that are used in default base/diff so as to provide a more graceful error message.


russellhadley:
@dotnet/jit-contrib

swaroop-sridhar:
:+1:

PR #30, opened 2016-08-23
Rewrite compile commands on Windows

adiaaida:
If the user did not specify a compile_commands.json, but one exists, we
will default to using it. This is to allow the use of
compile_commands.json files that were generated by cmake with NMake
Makefiles on Windows. These files require slightly more processing than
cmake files generated by cmake on non-windows platforms, as clang gets
confused by some of the windows specific flags, and we also have to force
the use of various other flags.

This change also modifies the output of DoClangFormat so that the output
doesn't get interleaved during the parallel runs of clang-format. We write
all the output to a list, and then print it out after we have run all
clang-format instances.

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -42,9 +42,8 @@ namespace ManagedCodeGen
             private bool _fix = false;
             private bool _verbose = false;
             private bool _ignoreErrors = false;
-            private string _buildLog = null;
             private string _compileCommands = null;
-            private bool _buildCompileCommands = true;
+            private bool _rewriteCompileCommands = false;
 
             private JObject _jObj;
             private string _jitUtilsRoot = null;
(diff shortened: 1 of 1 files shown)


adiaaida:
@RussKeldorph PTAL
/cc: @dotnet/jit-contrib

RussKeldorph:
>                     string logFullPath = Path.Combine(_rootPath, "bin", "Logs", logFile);

Do we still need build log processing after this change?

---
Refers to: src/jit-format/jit-format.cs:189 in e12f4a2. [](commit_id = e12f4a21521620ae3c7a66cd6714ce96d7981311, deletion_comment = False)

RussKeldorph:
LGTM

adiaaida:
I'd like to leave that functionality so that if people don't want to have to have build.cmd generate compile_commands.json and just use the build log to generate it, they can.

adiaaida:
@RussKeldorph I've updated this change to run build.cmd/sh if a compile_commands.json doesn't exist. I've removed some, but not all, of the logic where we check to see if a build log exists. Should I remove all of that logic, or give the user an option to build compile_commands.json from the log if they want to?

RussKeldorph:
@adiaaida I would prefer to remove the log processing logic unless there is a compelling case to keep it (is there?).  The only time I think it would be useful is for formatting "changed files only," but the confusion that introduces around what happens when a build has versus hasn't been run suggests to me that it's a bad idea.  For that general scenario, I would prefer that we add Git support (format staged files, format files affected by a commit, etc.).  To me that seems to address most or all of the interesting scenarios with way more robustness.  We don't have to do the Git thing for v1 though.

adiaaida:
:thumbsup:

adiaaida:
@RussKeldorph PTAL

src/jit-format/jit-format.cs
```diff
@@ -181,40 +180,61 @@ private void validate()
...
-                    }
-                    else
-                    {
-                        _buildLog = logFullPath;
-                        string[] compileCommandsPath = { _rootPath, "bin", "obj", "Windows_NT." + _arch + "." + _build, "compile_commands.json" };
+                        string[] compileCommandsPath = { _rootPath, "bin", "nmakeobj", "Windows_NT." + _arch + "." + _build, "compile_commands.json" };
                         _compileCommands = Path.Combine(compileCommandsPath);
+                        _rewriteCompileCommands = true;
+
+                        if (!File.Exists(_compileCommands))
```
RussKeldorph: I'm a little concerned about this being stale, but assuming it's not instantaneous to regenerate compile_commands, I'm ok with leaving this for now and seeing how it works.  Doesn't seem likely to cause a problem.

RussKeldorph:
LGTM

PR #36, opened 2016-08-30
Fix bugs in boostrap

adiaaida:
Bootstrap.sh calls validate_url, whose definition had parentheses. This
was not legal in the lab, so this change removes those parentheses.

Additionally, for the lab we need to install the 3.8 clang tools if they
are not yet installed. This change will download 3.8 if another version
of the tools is already installed.


adiaaida:
@RussKeldorph PTAL
/cc @dotnet/jit-contrib

RussKeldorph:
Source changes LGTM.  Before merging, please update the description to fully describe everything this change does, such as the version check.

PR #37, opened 2016-08-30
Update documentation for jit-format

adiaaida:
We have introduced a formatting python script for the lab that can also
be run locally that devs can use to format the jit directory. Update the
documentation to call it out.


adiaaida:
@RussKeldorph PTAL

RussKeldorph:
LGTM

PR #38, opened 2016-08-31
Create compile_commands directory if it doesn't exist

adiaaida:
If the obj\<os>.arch.build directory does not exist, we need to create
it or we will get a StackOverflowException. This should only happen on
Windows, as cmake will build into nmakeobj, not obj. CreateDirectory
will create the directory if it does not exist, and do nothing if the
directory already exists.


adiaaida:
@jashook PTAL

jashook:
lgtm

PR #42, opened 2016-09-01
Fix validate_url for bash and move tools download to function

adiaaida:
The back quotes for validate_url and the grep command were wrong for the
version of bash on our linux machines. Additionally, this change moves
the tools download to a function so we can call it both if we don't have
the tools installed and if we have the wrong version installed.


adiaaida:
@jashook PTAL

jashook:
lgtm

PR #46, opened 2016-09-26
Allow bootstrap.cmd to be used for building without doing a clone

BruceForstall:
After this there are two "modes": 1. invoke bootstrap.cmd from an existing clone, or 2. invoke it from a downloaded copy. With 1, it just builds everything (and downloads clang-format/clang-tidy, if necessary).

@adiaaida @russellhadley PTAL

### DIFF
--- a/bootstrap.cmd
+++ b/bootstrap.cmd
@@ -1,15 +1,98 @@
-:: Quick and dirty bootstrap. 
+@echo off
 
-where /q dotnet.exe || echo Can't find dotnet.exe! Please add to PATH && goto :EOF
-where /q git.exe || echo Can't find git.exe! Please add to PATH && goto :EOF
-set root=%~dp0
-pushd %root%
+REM Bootstrap the jitutils tools:
+REM 1. If this script is run not from within the jitutils directory (e.g., you've downloaded
+REM    a copy of this file directly), then "git clone" the jitutils project first. Otherwise,
+REM    if we can tell we're being run from within an existing "git clone", don't do that.
+REM 2. Build the jitutils tools.
+REM 3. Download (if necessary) clang-format.exe and clang-tidy.exe (used by the jit-format tool).
+
+where /q dotnet.exe
+if %errorlevel% NEQ 0 echo Can't find dotnet.exe! Please install this ^(e.g., from https://www.microsoft.com/net/core^) and add dotnet.exe to PATH && goto :EOF
+
+where /q git.exe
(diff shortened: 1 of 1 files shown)


bootstrap.cmd
```diff
@@ -1,16 +1,66 @@
...
+if %errorlevel% NEQ 0 echo Can't find dotnet.exe! Please install this ^(e.g., from https://www.microsoft.com/net/core^) and add dotnet.exe to PATH && goto :EOF
+
+set __root=%~dp0
+setlocal
+
+REM Are we already in the dotnet/jitutils repo? Or do we need to clone it?
+REM We look for sample_asmdiff.json and sample_format.json files in the current directory
+REM (which is the directory this script was invoked from).
+
+if not exist %__root%sample_asmdiff.json goto clone_jitutils
```
adiaaida: I feel like it would be better to run something like git remote -v and grep the output for jitutils.git, rather than check for the existence of a particular file. What if either of these files get removed from the repo?

BruceForstall: If the files are removed from the repro then yes, we would need to update bootstrap.cmd. Doesn't seem like a big deal.

The "git remote -v" solution wouldn't verify that the directory we're invoked from is also the root of the repo. I suppose I could only check for build.cmd, which must be there.

@adiaaida @russellhadley comments?

adiaaida: I'd be ok with looking for build.cmd. You can get the top level of the repo with this command:

git rev-parse --show-toplevel

Which you could compare to the current working directory.

BruceForstall:
@adiaaida ok, I updated the checks to only look for build.cmd. I also compare the directory with build.cmd against the root of a repo, but I'm not sure that's too useful (and it's complex), but, well, why not. Comments?

adiaaida:
The problem now is that if you're running bootstrap.cmd from, say, inside the coreclr repo, you're going to find build.cmd and you're going to be in the root of the coreclr repo, so it's going to think that it's in the jitutils repo. I think what I would do is this:

git remote -v > file.txt
findstr/grep jitutils file.txt // make sure that we are in the jitutils repo
gitroot = git rev-parse --show-toplevel
make sure %__root% == gitroot

BruceForstall:
@adiaaida Ok, I updated the change to check `git remote` also. I did it after the `git rev-parse` but I don't think the order matters. I also changed to always fall back to the original case of cloning the repo on any of these condition failures.

adiaaida:
LGTM.

PR #49, opened 2016-10-13
Factor diff logic, add support for --crossgen flag

sbomer:
This change moves all of the diff logic to jit-diff, which will invoke
jit-dasm once for each tag.

jit-dasm now takes an argument that specifies the crossgen executable to
use, and an optional jit argument that will be passed along to crossgen
via /JITPath.

jit-diff has been modified to take an additional argument that specifies
the crossgen executable to use. With this argument, jit-diff will use
clrjit from the the --base and --diff arguments (which are now
directories instead of the full crossgen path) with the specified
crossgen. This should prevent changes to
System.Private.CoreLib.dll (which must match the crossgen version) from
manifesting as jit diffs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,17 @@ Sample help commandlines:
                     [-a] [-t <arg>] [-c] [-f] [--benchmarksonly] [-v]
                     [--core_root <arg>] [--test_root <arg>]
     
-        -b, --base <arg>        The base compiler path or tag.
-        -d, --diff <arg>        The diff compiler path or tag.
-        --crossgen <arg>        The crossgen compiler exe.
+        -b, --base <arg>        The base compiler directory or tag. Will use
+                                crossgen or clrjit from this directory,
+                                depending on whether --crossgen is
+                                specified.
+        -d, --diff <arg>        The diff compiler directory or tag. Will use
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -78,18 +78,21 @@ Sample help commandlines:
...
-                    [-m] [-f] [-v] [--core_root <arg>] [--test_root <arg>]
-
-        -b, --base <arg>        The base compiler exe or tag.
-        -d, --diff <arg>        The diff compiler exe or tag.
+    usage: jit-diff diff [-b <arg>] [-d <arg>] [--crossgen <arg>] [-o <arg>]
+                    [-a] [-t <arg>] [-c] [-f] [--benchmarksonly] [-v]
+                    [--core_root <arg>] [--test_root <arg>]
+    
+        -b, --base <arg>        The base compiler path or tag.
+        -d, --diff <arg>        The diff compiler path or tag.
```
russellhadley: Need to make it clear that these are JIT dlls for the explict case.

sbomer: I added more details. Note that I've changed the base and diff to be the directories containing crossgen/clrjit, which is consistent with the config files.

PR #53, opened 2016-10-19
Avoid downloading clang-format and clang-tidy

BruceForstall:
If they've already been downloaded to the 'bin' directory, where
we download them, then don't download them again.

### DIFF
--- a/bootstrap.cmd
+++ b/bootstrap.cmd
@@ -102,6 +102,14 @@ dotnet restore
 
 call .\build.cmd -p -f
 
+:: Check to see if clang-format and clang-tidy are available. Since we're going
+:: to add the 'bin' directory to the path, and they most likely already live there
+:: if you're running bootstrap.cmd not for the first time, add 'bin' to the path
+:: here (within the setlocal scope for now) before checking for them. Add 'bin'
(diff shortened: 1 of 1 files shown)


BruceForstall:
@adiaaida @russellhadley PTAL

adiaaida:
LGTM

bootstrap.cmd
```diff
@@ -102,6 +102,14 @@ dotnet restore
 
 call .\build.cmd -p -f
 
+:: Check to see if clang-format and clang-tidy are available. Since we're going
+:: to add the 'bin' directory to the path, and they most likely already live there
+:: if you're running bootstrap.cmd not for the first time, add 'bin' to the path
+:: here (within the setlocal scope for now) before checking for them. Add 'bin'
+:: at the end of the path, to prefer other user downloaded versions, if any.
+
+set PATH=%PATH%;.\bin
```
BruceForstall: We don't know the absolute path, because this code is reached from two different locations, with different absolute paths. Later, when 'bin' is added to the path "for real", the absolute path is used.

adiaaida:
LGTM

PR #56, opened 2016-10-25
Work around command length restriction

sbomer:
Looks like the maximum command length is severely restricted by a special case in the command resolution logic that only kicks in for .cmd files: [WindowsExePreferredCommandSpecFactory.cs](https://github.com/dotnet/cli/blob/3a567e595773b61650f2dcd1c7ec2b53b91cd7a8/src/Microsoft.DotNet.Cli.Utils/CommandResolution/WindowsExePreferredCommandSpecFactory.cs). Changing the wrapper extension to .bat works around this. There will still be a max length of ~8000 characters, and if we ever run into that, I can come up with a more robust solution (such as starting a process for each library passed as an argument, or batching them up).


russellhadley:
LGTM

PR #57, opened 2016-10-25
Don't print message about missing JIT_UTILS_ROOT

adiaaida:
We don't want to message about missing JIT_UTILS_ROOT. Instead, we will
message when we find a JIT_UTILS_ROOT and inform the user that a
configuration was loaded.


adiaaida:
@RussKeldorph PTAL
/cc: @dotnet/jit-contrib

PR #58, opened 2016-10-25
Change output for clang-format

adiaaida:
When clang-format reports that there are formatting errors, rather than
printing that the user should rerun in fix mode, give a more explicit
command line.

Additionally, print the first 10 formatting errors when not in verbose
mode.

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -462,8 +462,24 @@ namespace ManagedCodeGen
             {
                 if (!RunClangFormat(filenames, config.Fix, verbose))
                 {
-                    Console.WriteLine("Clang-Format needs to be rerun in fix mode");
-                    returncode = -1;
+                    Console.WriteLine("Clang-format found formatting errors.");
+                    returncode -= 2;
+                }
+            }
+
+            if (returncode != 0)
+            {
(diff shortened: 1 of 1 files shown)


adiaaida:
@RussKeldorph PTAL
/cc: @dotnet/jit-contrib

src/jit-format/jit-format.cs
```diff
@@ -464,7 +464,15 @@ public static int Main(string[] args)
             {
                 if (!RunClangFormat(filenames, config.Fix, verbose))
                 {
-                    Console.WriteLine("Clang-Format needs to be rerun in fix mode");
+                    Console.WriteLine("Clang-format found formatting errors. Run:");
+                    if (returncode == -1)
```
RussKeldorph: The analogous tidy case above seems to need a fix as well.  Also, you have to run without -noformat to get here, right?  I think that you need to move this error handling out to the main path and handle the various combinations of error conditions, especially if you're going to tailor the suggested command line with -untidy/-noformat based on the situation.

adiaaida: You'd never run with --noformat when you run --fix with clang-tidy, as clang-tidy doesn't correctly format the code it changes. So the options are --fix or --fix --untidy.

adiaaida:
@RussKeldorph PTAL

PR #69, opened 2016-11-10
Add support to cijobs for retrieving builds by commit

sbomer:
This change adds options to the list and copy commands that allow
retrieving the information or artifacts for a specific commit, and
removes compiler warnings about unused variables and uninitialized
fields that are only set by json deserialization.


russellhadley:
LGTM

PR #70, opened 2016-11-14
Update framework assembly list, validate directory args

sbomer:
This change updates the built-in framework assembly list adds some validation to check that the base, diff, and output directories passed in as arguments exist.

With the larger list of framework assemblies, the call to jit-dasm was hitting the command length limit, so this change will also run jit-dasm on each assembly individually.


src/jit-diff/jit-diff.cs
```diff
@@ -547,50 +563,100 @@ public int List()
...
+            "System.Threading.Thread.dll",
+            "System.Threading.ThreadPool.dll",
+            "System.Threading.Timer.dll",
             "System.Xml.ReaderWriter.dll",
             "System.Xml.XDocument.dll",
-            "System.Xml.XmlDocument.dll"
+            "System.Xml.XmlDocument.dll",
+            "System.Xml.XPath.dll",
+            "System.Xml.XPath.XDocument.dll"
         };
```
russellhadley: When you run these new assemblies through the diff do they produce results?  There were a set of assemblies that were just type forwarders that I omitted in the first round.  Just curious if you did the filtering.

sbomer: Good question. The following assemblies don't have any dasm output:

mscorlib.dasm
System.AppContext.dasm
System.Globalization.Calendars.dasm
System.Globalization.dasm
System.IO.FileSystem.Primitives.dasm
System.Reflection.dasm
System.Reflection.Emit.dasm
System.Reflection.Emit.ILGeneration.dasm
System.Reflection.Emit.Lightweight.dasm
System.Reflection.Primitives.dasm
System.Resources.ResourceManager.dasm
System.Runtime.Loader.dasm
System.Security.Principal.dasm
System.Text.Encoding.dasm
System.Threading.Thread.dasm
System.Threading.ThreadPool.dasm
System.Threading.Timer.dasm

I suppose I'll remove them from the list.

Issue #68, opened 2016-11-07
Investigate “error: Can't find --crossgen tool.”


sbomer:
This error was showing up when running the command:
C:\git\tools\jitutils>bin\jit-diff.cmd diff --analyze --frameworksonly --crossgen C:/git/coreclr/bin/Product/Windows_NT.x64/crossgen  --output C:/git/out --core_root C:/git/tools/jitutils/fx --base C:/git/coreclr/bin/Product/Windows_NT.x64.Debug --diff C:/git/coreclr/bin/Product/Windows_NT.x64.Debug_base

Issue #67, opened 2016-11-07
Jit-diff should look for specific name of dlls instead of globbing


sbomer:
Currently jit-diff looks for "\*clrjit\*" as the jit tool, which causes trouble when there are two files in the base or diff directory matching that pattern. Jit-diff should only look for files named exactly clrjit.dll (libclrjit.so on linux). The same fix should be made for crossgen as well.

PR #73, opened 2016-11-15
Look for exact name of clrjit and crossgen

sbomer:
This change fixes [67](https://github.com/dotnet/jitutils/issues/67) and [68](https://github.com/dotnet/jitutils/issues/68).

Instead of using pattern matching to find clrjit and crossgen, we now look for the exact names of these binaries: libclrjit.so/crossgen on linux, and clrjit.dll/crossgen.exe on windows.

The documentation update reminds the user to use the platform-specific crossgen executable name.

### DIFF
--- a/doc/getstarted.md
+++ b/doc/getstarted.md
@@ -122,7 +122,7 @@ Steps:
   from the baseline CoreCLR build to \<jitutils_repo\>/fx (overwriting existing versions).
 * Invoke command
 ```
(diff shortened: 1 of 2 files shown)


sbomer:
@russellhadley PTAL

russellhadley:
LGTM but can you rebase so it's one commit rather than the big list you have?

PR #76, opened 2016-12-09
Don't emit error files from jit-dasm for non-managed assemblies

sbomer:
Also fix dependency versions to be consistent between tools and add
version numbers used to publish packages for the coreclr jitdiff jobs


sbomer:
@russellhadley

russellhadley:
LGTM

PR #78, opened 2017-01-09
Make jit-format use the same version of packages as the other tools

sbomer:
Jit-format was running into issues after a change in the dependency
versions of other jitutils tools.

Because we publish all of the jitutils tools into a common directory,
the dependencies of the last package to get published will end up in
the publish directory. Eventually we should fix this behavior, but for
now modifying jit-format to use the same dependency versions as cijobs
works around the issue.

@adiaaida @russellhadley

### DIFF
--- a/src/jit-format/project.json
+++ b/src/jit-format/project.json
@@ -6,11 +6,12 @@
     },
 
     "dependencies": {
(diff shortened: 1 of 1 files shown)


adiaaida:
LGTM.

PR #79, opened 2017-02-08
Upgrade for new cli versions

JosephTremoulet:
Migrate from project.json to csproj (works building/publishing with rc3,
should set things up favorably going forward)

Also update project references to cli-utils to use rc4, and rename
`PathUtility.EnsureParent` to `PathUtility.EnsureParentExists`
accordingly.


JosephTremoulet:
@russellhadley PTAL.
/cc @dotnet/jit-contrib, @AndyAyersMS 

@adiaaida / @jashook, what's a good way to validate that this won't hork the formatting jobs in the coreclr repo?  The CI machines would need to be using dotnet version [rc3](https://github.com/dotnet/core/blob/master/release-notes/rc3-download.md) or later.

adiaaida:
You could run the format.py script that's in the coreclr repo, except you'll want to bring in your own version of the jitutils repo, and edit the script so it doesn't try to download it.

adiaaida:
format.py command line:

python -u tests\scripts\format.py -c %WORKSPACE% -o Windows_NT -a x64

JosephTremoulet:
@adiaaida I've verified that the same "dotnet restore" and "build.cmd -p -f" commands in the jitutils repo still work, and that invoking thusly-built jit-diff from a coreclr repo also works.  All of that with dotnet-cli rc3 installed on my machine and in my path.  What I'm wondering is if the CI machines have a compatible version installed and on their path.

adiaaida:
The lab downloads a specific version from the .net cli website in the format.py script. We might need to upgrade the script to make sure they grab the right version, but we need to be able to download zip/tar files so that the script can install them.

JosephTremoulet:
Got it, thanks.  I've followed your steps and confirmed that the preview2 version that format.py currently pulls down is incompatible with this, and that updating format.py to the links from the rc3 release notes gets past that, but then discovered that jit-format fails with these changes, so I'm looking into that...

adiaaida:
Let me know if you want me to take a look at anything.

JosephTremoulet:
If you (or anyone) are familiar with the cli-command-utils stuff, here's the error with stack trace:

```
Unhandled Exception: System.InvalidOperationException: Specify which project file to use because this 'D:\Source\coreclr' contains more than one project file.
   at Microsoft.DotNet.Cli.Utils.ProjectFactory.GetMSBuildProjPath(String projectDirectory)
   at Microsoft.DotNet.Cli.Utils.ProjectFactory.GetMSBuildProj(String projectDirectory, NuGetFramework framework, String configuration, String outputPath)
   at Microsoft.DotNet.Cli.Utils.ProjectToolsCommandResolver.ResolveFromProjectTools(CommandResolverArguments commandResolverArguments)
   at Microsoft.DotNet.Cli.Utils.ProjectToolsCommandResolver.Resolve(CommandResolverArguments commandResolverArguments)
   at Microsoft.DotNet.Cli.Utils.CompositeCommandResolver.Resolve(CommandResolverArguments commandResolverArguments)
   at Microsoft.DotNet.Cli.Utils.Command.Create(ICommandResolverPolicy commandResolverPolicy, String commandName, IEnumerable`1 args, NuGetFramework framework, String configuration, String outputPath, String applicationName)
   at ManagedCodeGen.jitformat.TryCommand(String name, IEnumerable`1 commandArgs, Boolean capture)
   at ManagedCodeGen.jitformat.DoClangTidyInnerLoop(Boolean fix, Boolean ignoreErrors, String checks, String compileCommands, String filename, Boolean verbose)
```

We're trying to execute clang-tidy using a call to `Command.Create()`, which appears to be picking up the `DefaultCommandResolverPolicy`, which gives a `CompositeCommandResolver` that includes a `ProjectToolsCommandResolver` that fails. I'm guessing we'd want the `ProjectToolsCommandResolver` to just fail gracefully so that a subsequent resolver could find the command.  I tried manually specifying a `PathCommandResolverPolicy` (since we want to pick up clang-tidy from the path), but either misunderstood or did that incorrectly, because that doesn't manage to find it (I get `Error: clang-tidy command not found!  Add clang-tidy to the path.`).

JosephTremoulet:
Ok, it turns out that using the `PathCommandResolverPolicy` works fine and my issue was pilot error (clang-format actually wasn't in my path...).  So I've updated this PR to use the path resolver, and pushed up the coreclr changes in dotnet/coreclr#9437 that will make format.py work with this.

JosephTremoulet:
@dotnet/jit-contrib FYI building jit-format/jit-diff etc will require dotnet rc3 (which you can install from [here](https://github.com/dotnet/core/blob/master/release-notes/rc3-download.md)) or later once you've merged this change.

src/jit-diff/jit-diff.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <AssemblyName>jit-diff</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <PackageId>jit-diff</PackageId>
+    <PackageTargetFallback>$(PackageTargetFallback);dnxcore50</PackageTargetFallback>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Runtime.Serialization.Primitives" Version="4.1.1" />
+    <PackageReference Include="Microsoft.DotNet.Cli.Utils" Version="1.0.0-rc4-*" />
+    <PackageReference Include="System.CommandLine" Version="0.1.0-*" />
```
mikedn: While you're at it you could ensure that jit utils projects reference the same assembly versions. Otherwise you can end up with crashes like in #80

JosephTremoulet: Yep, thanks, I'd done that with Microsoft.DotNet.Cli.Utils but missed System.CommandLine.  I've updated to make all references consistent across all projects.

src/jit-format/jit-format.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <DebugType>portable</DebugType>
+    <AssemblyName>jit-format</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <PackageId>jit-format</PackageId>
+    <PackageTargetFallback>$(PackageTargetFallback);dnxcore50</PackageTargetFallback>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.Cli.Utils" Version="1.0.0-rc4-*" />
```
russellhadley: Should we go ahead and get rid of the '*' and pin to a particular version?

JosephTremoulet: Sure.  Updated.

BruceForstall:
@JosephTremoulet @russellhadley @sbomer Could we make jitutils not depend on ambient dotnet CLI but instead download whatever CLI it wants, like our dotnet/coreclr build does?

@JosephTremoulet If there's a new min version, you should update https://github.com/dotnet/jitutils/blob/master/doc/getstarted.md to state that. btw, I sync'ed jitutils and have `1.0.0-preview2-003121` on Windows and `bootstrap.cmd` seems to work, albeit with warnings:
```
D:\gh\jitutils\src\cijobs\project.json(12,30): warning NU1012: Dependency conflict. Microsoft.DotNet.ProjectModel 1.0.0-rc3-003121 expected Newtonsoft.Json >= 9.0.1 but received 8.0.3
D:\gh\jitutils\src\cijobs\project.json(12,30): warning NU1012: Dependency conflict. Microsoft.Extensions.DependencyModel 1.0.0 expected Newtonsoft.Json >= 9.0.1 but received 8.0.3
D:\gh\jitutils\src\cijobs\project.json(9,67): warning NU1012: Dependency conflict. Microsoft.DotNet.ProjectModel 1.0.0-rc3-003121 expected System.Runtime.Serialization.Primitives >= 4.1.1 but received 4.1.1-rc3-24210-10
```

russellhadley:
@BruceForstall that doesn't seem unreasonable.  Ideally we'd bake the whole thing together with the linker/bundler into a single exe with no other dependencies, but that's still in the future.  (@swaroop-sridhar could we make the jit-utils tools one of our bundler test cases?)  Put it in the issues and maybe someone can opportunistically grab it, otherwise it will be on the list once I can free up a resource to get back to the lab diff work.

sbomer:
If I understand correctly I think there are two distinct asks:
@BruceForstall would like the jitutils **build** to download and use a specific version of the dotnet cli.
@russellhadley is referring to another issue where the dependencies of the jitutils need to match some dependencies of our corecelr tests, because the jit-dasm jobs pulls in jitutils packages and their dependencies into a test runtime directory.

PR #82, opened 2017-02-10
Update to use dotnet cli rc3 (again)

JosephTremoulet:
Restore the changes from #79, plus two fixes:

 1. Adjust build.sh to compute the install directory differently,
    since the path passed to dotnet publish is now taken to be
    relative to the project, not relative to the current path; use
    $PWD to get an absolute path to pass.
 2. Adjust bootstrap.sh to compute the RID differently; we get it
    by munging the `dotnet --info` output, which now includes an
    additional line after the RID line.


JosephTremoulet:
@russellhadley / @adiaaida PTAL

JosephTremoulet:
I should mention that I verified locally that running format.py (as updated in dotnet/coreclr#9437) works on Ubuntu.

adiaaida:
LGTM

PR #84, opened 2017-02-13
Use ScriptCommandResolverPolicy

JosephTremoulet:
The `DefaultCommandResolverPolicy` includes a `ProjectToolsCommandResolver`
which chokes if it finds multiple project files, so e.g. running jit-diff
from the root coreclr directory currently fails.  Update all the calls to
`Command.Create` to instead use `ScriptCommandResolverPolicy` via a local
wrapper class to proxy the interface call to the static `Create` method.

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -553,6 +553,11 @@ namespace ManagedCodeGen
             }
         }
 
+        class ScriptResolverPolicyWrapper : ICommandResolverPolicy
+        {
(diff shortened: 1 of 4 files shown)


JosephTremoulet:
@adiaaida / @russellhadley PTAL.  I've validated this against format.py on Linux and Windows using the same technique as #79.  Also verified that I can successfully run jit-diff from the root coreclr directory with this, and that jit-diff can find e.g. jit-dasm with this without requiring it to be on the path.

adiaaida:
Looks good to me.

PR #85, opened 2017-02-14
Update getstarted.md with cli version info

JosephTremoulet:
Point out that we require a csproj-compatible version of the dotnet cli
tools.


JosephTremoulet:
@BruceForstall PTAL

BruceForstall:
LGTM (although I prefer non-slang "understand" to "grok")

JosephTremoulet:
Updated per suggestion.

PR #86, opened 2017-02-15
Clean up generated native files

### LINKED ISSUE #74: Jit-dasm should clean up generated .ni.exe files
Failure to do so leaves them in the test directories (if you were diffing tests), which means that from then on when you run tests you'll be picking up the crossgened code and not testing what you intended.

JosephTremoulet:
Update jit-disasm to delete the .ni.exe/dll files it creates, to make sure
they don't unintentionally get picked up by subsequent test runs in the
same test tree.

Resolves #74


JosephTremoulet:
@dotnet/jit-contrib PTAL

PR #90, opened 2017-02-23
Fix `build -f`

JosephTremoulet:
In the migration to csproj, the frameworks-populating of `build -f` was
lost, because packages.csproj can't specify standalone like project.json
could.  With csproj, an explicit --runtime argument must be passed to
dotnet restore in order to publish as a standalone.  The dotnet publish
--runtime in turn requires that the project has been restored for the
specified runtime, so add an explicit `dotnet restore --runtime` in the
build scripts when processing -f.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -16,7 +16,6 @@ set buildType=Release
 set publish=false
 set fx=false
 
-REM REVIEW: 'platform' is never used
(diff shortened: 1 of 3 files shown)


JosephTremoulet:
@BruceForstall / @russellhadley / @sbomer PTAL.

sbomer:
LGTM

BruceForstall:
As long as it works, LGTM

PR #95, opened 2017-03-17
Finish all diff tasks; don't exit early on failure

BruceForstall:
If an asm run fails, e.g., in the baseline run, we were
immediately exiting. Instead, run all the baseline, diff,
and analyze tasks, so you can get as much information as
possible from the code you've got.

### DIFF
--- a/src/jit-diff/jit-diff.cs
+++ b/src/jit-diff/jit-diff.cs
@@ -995,29 +995,29 @@ namespace ManagedCodeGen
                 }
             }
 
+            int status = 0;
+
             if (config.HasBasePath)
             {
-                int status = RunDasmTool(config, commandArgs, assemblyArgs, "base", config.BasePath);
-
-                if (status != 0)
+                int baseStatus = RunDasmTool(config, commandArgs, assemblyArgs, "base", config.BasePath);
+                if (baseStatus != 0)
                 {
-                    return status;
+                    status = baseStatus;
                 }
             }
 
             if (config.HasDiffPath)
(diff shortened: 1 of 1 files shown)


BruceForstall:

JosephTremoulet:
Thanks for working on this.  When I run jit-diff, there are a gazillion lines of output spew, then it tells me how many failures there are.  Typically (when diffing the full tree), I do this for base, note the number of failures, then do this for diff and check that the failure count matches (and yes it would be better if there weren't failures, it just hasn't become a high priority to dig into them).  With this change as-is, would the note about how many failures there were in the base set be buried above the gazillion lines of output spew from the diff run?  If so, any way to avoid that?

JosephTremoulet:
For the purpose of comparing failure counts, I could always just keep my same workflow.  So I think the more important point is: when one of the first steps fail but the last step doesn't, will that be obvious from the tail of the output, or would we have to grep the gazillion-line spew to determine it?

BruceForstall:
> With this change as-is, would the note about how many failures there were in the base set be buried above the gazillion lines of output spew from the diff run? If so, any way to avoid that?

Yes, it would be buried (in my particular example, it is buried).

We could easily print/output the error counts at the very end.

JosephTremoulet:
> We could easily print/output the error counts at the very end.

SGTM

russellhadley:
Printing out the error counts at the end would be great.  Other than that LGTM.  Thanks for looking at this.

PR #96, opened 2017-03-18
Fix bootstrap.sh to only clone jitutils when necessary

BruceForstall:
If bootstrap.sh is invoked from the jitutils repo root directory,
then the build is done in the existing repo. Otherwise, the jitutils
repo is cloned and then built.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,64 +1,148 @@
-#Quick and dirty bootstrap. 
+#!/usr/bin/env bash
+
+# Bootstrap the jitutils tools:
+# 1. If this script is run not from within the jitutils directory (e.g., you've downloaded
+#    a copy of this file directly), then "git clone" the jitutils project first. Otherwise,
+#    if we can tell we're being run from within an existing "git clone", don't do that.
+# 2. Build the jitutils tools.
+# 3. Download (if necessary) clang-format and clang-tidy (used by the jit-format tool).
+
+function get_host_os {
+    # Use uname to determine what the OS is.
+    OSName=$(uname -s)
+    case $OSName in
+        Linux)
+            __HostOS=Linux
+            ;;
+
+        Darwin)
+            __HostOS=OSX
+            ;;
+
+        FreeBSD)
+            __HostOS=FreeBSD
+            ;;
+
+        OpenBSD)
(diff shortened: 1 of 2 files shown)


BruceForstall:

bootstrap.sh
```diff
@@ -1,77 +1,159 @@
...
-pushd ./jitutils
+if [ ${__clone_repo} == 1 ]; then
+    git clone https://github.com/dotnet/jitutils.git
+    pushd ${__root}/jitutils >/dev/null
+else
+    pushd . >/dev/null
+fi
 
-# Pull in needed packages.  This works globally. (due to global.json)
+# Pull in needed packages.  This works globally (due to global.json).
```
JosephTremoulet: Looks like I missed this comment when migrating to csproj. There isn't a global.json  anymore, it's due to jitutils.sln now.

russellhadley:
Looks fine to me too.  Bootstrap appears to be morphing into a "get jitutils up to date on my machine" rather than the first time install script I stuck in originally.  I'm wondering if we should change the name to "install*" and wrap the whole thing up conceptually as install -update.  (which could be the default).  Bootstrap was intended to be a bridge as people got familiar with the repo but now that there's extra bits for clang-format/-tidy it seems like it's seems like a more central kind of installer/updater.  I'm probably over thinking this.  Feel free to merge as is, if the name change is motivating/persuasive though, I'm fine with that change too.

adiaaida:
This looks good to me.

If you do decide to go the name change route, I'd probably try to get it in on a weekend when there are fewer checkins so you can get changes in jitutils and coreclr (for the formatting legs in) in at the same time when no one could potentially start failing them because of file name changes.

BruceForstall:
>  Bootstrap appears to be morphing into a "get jitutils up to date on my machine" 

Yes. Before the clang-format/clang-tidy downloading, it was less so, and doing the individual steps were less onerous to be done individually. But I like the "clone; bootstrap; run" simplicity. Sure, the name doesn't quite match the functionality anymore.

BruceForstall:
> You're porting functionality that's already in build.cmd, yes?

Bootstrap.cmd; yes. They should be in sync now.

PR #97, opened 2017-03-18
Fix comment about global.json

BruceForstall:


BruceForstall:
@JosephTremoulet PTAL

Issue #80, opened 2017-02-09
Running `jit-diff diff` crashes


BruceForstall:
```
c:\gh\coreclr>jit-diff diff
Environment variable JIT_UTILS_ROOT not found - no configuration loaded.

Unhandled Exception: System.MissingMethodException: Method not found: 'Microsoft.DotNet.Cli.Utils.Command Microsoft.DotNet.Cli.Utils.Command.Create(System.String, System.Collections.Generic.IEnumerable`1<System.String>, NuGet.Frameworks.NuGetFramework)'.
   at ManagedCodeGen.jitdiff.TryCommand(String name, IEnumerable`1 commandArgs, Boolean capture)
   at ManagedCodeGen.jitdiff.Config.SetRID()
   at ManagedCodeGen.jitdiff.Config..ctor(String[] args)
   at ManagedCodeGen.jitdiff.Main(String[] args)
```

Maybe this should output the "help" screen, but it shouldn't crash.

mikedn:
I think I ran into this before. The project.json files used by jit utils reference different versions of some assemblies and all projects output to the same bin directory. So we end up with executables that were compiled against some assembly version X running against assembly version Y.

PR #98, opened 2017-03-20
Reconstruct sln/csproj files in VS2017

BruceForstall:
There is some magic here that allows VS2017 to allow better
debugging, e.g., setting of command-line arguments.

### DIFF
--- a/jitutils.sln
+++ b/jitutils.sln
@@ -1,101 +1,52 @@
-
+﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.26228.9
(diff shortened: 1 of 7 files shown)


BruceForstall:

JosephTremoulet:
- You have different versions of `System.CommandLine` referenced in different csproj's; that needs to be fixed to avoid running into issues like #80 again.
 - The first line in each csproj looks the same to me, but gets highlighted in the github diff view.  Is there a BOM or a non-ascii quote or different line ending or something?  I guess it doesn't matter as long as restore/build/publish still work on all platforms.

BruceForstall:
There is a BOM.

I'll check the versions. It was a very manual process trying to get all csproj's to look the same.

BruceForstall:
@JosephTremoulet updated. Also, tested on Mac OSX and Ubuntu 16.04 -- seems to work.

Issue #75, opened 2016-11-17
Jit-dasm should use /fragile crossgen


JosephTremoulet:
Jit-dasm uses crossgen to avoid having to execute the code, but crossgen has several limitations compared to jitting code.  Crossgen has some sort of /fragile option to lift those constraints, and using this mode would mean the generated diffs would be more representative of diffs that will arise in jitted code.  Presumably it will be interesting to be able to select whether /fragile is desired or not when generating disasm/diffs, and I think that /fragile should be the default.

JosephTremoulet:
It's easy enough to plumb through --fragile (a4ae43b2b68f), but that turns out to be insufficient because /FragileNonVersionable crossgen requires dependencies to be already crossgened so it can find their native image files.  This would require making sure frameworks are crossgened if we want to crossgen benchmarks, making sure frameworks are crossgen'ed in the right order, and doing something about the fact that we currently delete each native image file immediately after generating it (which isn't as straightforward as you'd think because jit-dasm takes responsibility for deleting those files, but jit-diff invokes jit-dasm separately for each assembly [at least for frameworks assemblies], and trying to change that gives "command too long" errors so would need to use response files or something...).

JosephTremoulet:
>  I think that /fragile should be the default

FWIW, I've changed my mind about this as well.  I think it makes sense to look at diffs of ReadyToRun frameworks since that's what ships.  For benchmarks, it would make the most sense to look at jit codegen, but /FragileNonVersionable is something in between ReadyToRun and Jit.  Though /FragileNonVersionable is still likely better for benchmarks than ReadyToRun.

russellhadley:
@JosephTremoulet #99 linking back here for JIT like codegen. 

- I'd like to see more JIT like code gen in crossgen for testing purposes.  For the new optimization levels on your backlog I'd want to use crossgen for stress and "random" seed testing to try and generate more coverage of tiering while maintaining stable repros.
- I'd expect that we would then harness our big scenarios for the above and then be able to do multi optimization level diffs using Azure resources.
- Additionally it'd be super nice to use a pseudo random IBC data input to stress codegen as we make more optimizations in the JIT sensitive to it.

So yes more JIT-like modes and more ability to model/force tiering like execution.  Given how much easier it is to debug/validate crossgen cases, bringing it closer to our JIT codegen, and peak codegen can only help.

[...]

PR #99, opened 2017-03-21
Update jit-diff operation order, and parallelize

BruceForstall:
I reordered running diffs so base and diff of a single assembly are done back-to-back, so you can do diffs while the rest of the assemblies continue to be worked on.

The parallel work is in progress.

Any comments/suggestions?

### DIFF
--- a/src/jit-diff/jit-diff.cs
+++ b/src/jit-diff/jit-diff.cs
@@ -848,7 +848,7 @@ namespace ManagedCodeGen
             }
 
             private Config m_config;
-            private ConcurrentQueue<DasmWorkItem> DasmWorkQueue = new ConcurrentQueue<DasmWorkItem>();
+            private List<Task<int>> DasmWorkTasks = new List<Task<int>>();
 
             public DiffTool(Config config)
             {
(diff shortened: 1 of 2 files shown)


BruceForstall:
@russellhadley @JosephTremoulet @mikedn Comments?
cc @dotnet/jit-contrib

JosephTremoulet:
Thanks again for working on this stuff.

I've been taking a stab at #75 and running into the issue that running crossgen with /FragileNonVersionable on one assembly may demand that it can find the output of having run it on other assemblies that this one depends on (haven't gotten far enough to see if it expects all dependencies to be crossgen'ed, or if there's just a fixed set like S.P.CoreLib and friends -- what are ngen's requirements?).  That probably would have been more straightforward to work out without this interleaving.  Also, for parallelizing, will it be an issue having base and diff race to write the same output?

What I'm getting at is wondering what we'll need to do in terms of managing output .ni.exe/dll files in order to get interleaved/parallel/fragile all working.  If the answer is "for now we'll use some sort of locking in parallel to avoid races and /fragile is TBD", that's fine, I just wanted to surface the question and see if someone has a better answer (e.g. should we copy the input assemblies to some intermediate/output directory before running crossgen on them?).

[also, and this is probably a discussion that should happen in #75 rather than here, I'm wondering more broadly what we should do to get jit-like codegen -- things like "how close is /fragile to jit?" and "should it become closer to jit?" and "what's in the way of us using spmi for our typical asm diff needs?"]

BruceForstall:
We've never done parallel asm before, so that's new territory: on desktop with ngen-based asm diffs, ngen enforces serialization. On desktop, we very carefully laid out the list of assemblies to generate so dependencies were done before the assembly itself is done.

As for interleaving, presumably they should never write the same output files.

BruceForstall:
What is the rule with crossgen of Ready2Run assemblies? Are they all independent, so crossgen order doesn't matter? Or do you have to do System.Private.Corelib.ni.dll first (as fragile, always?) and then everything after that is independent?

src/jit-diff/jit-diff.cs
```diff
@@ -830,239 +835,330 @@ public static int InstallCommand(Config config)
...
+                int dasmFailures = 0;
 
-                var clrjitPath = FindJitLibrary(clrPath);
-                if (clrjitPath == null)
+                while (true)
                 {
-                    Console.Error.WriteLine("clrjit not found in " + clrPath);
-                    return -1;
+                    DasmWorkItem item;
+                    if (DasmWorkQueue.TryDequeue(out item))
```
mikedn: FWIW I've tried something much simpler for parallelization - simply start the original `RunDasmTool` using `Task.Run` for both base and diff. This avoids ordering issues and hogging the CPU. Granted, it doesn't make things faster if you don't actually run a base/diff session (AFAIR it's possible to run just the diff part but I've never tried that myself so I don't know).

BruceForstall:
That final line:
```
Native image c:\diffs\13\base\Microsoft.CSharp.ni.dll generated successfully.
```
is quite annoying, for creating diffs.

The Zapper does have a variable for silencing this, but it doesn't appear it's possible to enable that right now:
```
    if (!m_pOpt->m_silent)
    {
        GetSvcLogger()->Printf(W("Native image %s generated successfully.\n"), strNativeImagePath.GetUnicode());
    }
```

Nothing a COMPlus variable couldn't fix...

JosephTremoulet:
> delete the last line of the .dasm file which contains the .ni file path. But I wonder if that lines is always the last...

We could filter the output through a regex that replaces the path with a fixed string?

mikedn:
> We could filter the output through a regex that replaces the path with a fixed string?

I was planning to try something like that. The main trouble is that we probably need to implement a custom `TextWriter` that does this or deal with process output redirection ourselves instead of using the existing utilities.

russellhadley:
@BruceForstall would fixing the ni path image be as simple as adding an --silent option to crossgen?  Since we're invoking via command line anyway...

JosephTremoulet:
> we probably need to implement a custom TextWriter that does this or deal with process output redirection ourselves

Ah, that makes sense (though if stripping the last line doesn't have the same issue, I'm not following why not).  FWIW, it wouldn't be entirely unreasonable to have a separate jit-sed utility that normalizes dasm output, although:
 1. We're internally already doing the rest of what a jit-sed would do, via COMPlus_JitDiffableDisasm
 2. I haven't actually looked at the APIs involved here to know if piping through a separate process would be any simpler than the alternatives mentioned above.

src/jit-diff/jit-diff.cs
JosephTremoulet: > (AFAIR it's possible to run just the diff part but I've never tried that myself so I don't know

You're recalling correctly, and that's what I've been doing, both to reconcile error counts (which is now moot) and because it lets me be lazy and always invoke jit-diff through a wrapper script that fills in all the paths for me (assuming the jit in bin\Product\blah-Checked is what I want) so I can just build+gen-diff+build-master+gen-base rather than keeping track of where I'm putting different jit dlls.

Which is just a long-winded way of saying that I for one would appreciate making things faster even if I don't run base/diff together, although some sort of --max-threads option to avoid hogging the CPU would be appreciated too.

BruceForstall:
> would fixing the ni path image be as simple as adding an --silent option to crossgen? Since we're invoking via command line anyway...

Yes, which is what Jan Kotas suggested to me as well.

BruceForstall:
PR to add crossgen "/silent" option: https://github.com/dotnet/coreclr/pull/10350

BruceForstall:
@JosephTremoulet @russellhadley This is ready to be checked in. PTAL.

JosephTremoulet:
I'd prefer to keep reporting base failure count and diff failure count separately, rather than their sum, but otherwise looks good to me.

russellhadley:
LGTM

BruceForstall:
> I'd prefer to keep reporting base failure count and diff failure count separately, rather than their sum, but otherwise looks good to me.

It's not so easy now since base and diff tasks are interleaved. We can put it on the "to do" list. Note that it's easy to see by looking at the *.err files that are created.

PR #100, opened 2017-03-21
Pass /silent to crossgen

BruceForstall:
This allows base/diff generated dasm files to be identical,
even if the .ni.dll was generated to different paths.


BruceForstall:
@JosephTremoulet @russellhadley PTAL

PR #103, opened 2017-03-26
Improve jitutils

BruceForstall:
Short summary:

1. Improved usage model for jit-diff, especially, add support for much simpler
   command lines using better defaults. Read the rewritten diffs.md
   for details.
2. Improved jit-diff parallelism when diffing the test tree. (Previously,
   dasm generation within a recursive directory walk was not parallel.)
3. Changed jit-diff argument parsing to output the usage message on (most)
   errors (namely, the errors we validate, not the errors the argument
   library validates. It would be nice to do both.).
4. Fixed many crashing bugs

Details:

Using jit-diff is cumbersome due to the need to specify so many
long paths. These can be specified in the config.json file in
the directory pointed to by the `JIT_UTILS_ROOT` environment
variable, but that is inflexible.

This change introduces an alternate way to specify all the required
data for jit-diff, by detecting or specifying the dotnet/coreclr
repo root, and automatically finding all the paths that are required.

The defaults can be altered by using the new `--base_root`, `--diff_root`,
`--arch`, and `--build` arguments, or their respective config.json settings.

See the rewritten diffs.md document for much more detail and various
examples of using this.

Also:
1. To generate asm diffs for the test tree, specify the new "--tests"
argument, as well as the "--test_root" argument (if "--test_root" can't
be determined using the defaults).
2. jit-diff now automatically runs jit-analyze. It now has a "--noanalyze"
option to prevent this.
3. jit-diff now takes an optional "--gcinfo" argument (which it passes
through to jit-dasm) to specify emitting GC info into the generated dasm
files.
4. jit-format automatically detects the repo root, if run with the
current directory within the dotnet/coreclr repo. So, the "--core_root"
argument is no longer required, if the repo root can be found.
5. Rewrote "getting started" document to account for new defaults,
and other changes.
6. Move all cijobs documentation to cijobs.md.
7. Remove `--tag` option from `jit-dasm`: simply pass the `--output` path you desire.
8. Support diff'ing System.Private.CoreLib.dll or the frameworks assemblies, in addition to
   the benchmarks or the full test tree. I regularized this, and added a lot of missing
   argument checking. I also renamed the arguments since "only" is incorrect. So, now
   there is `--corelib`, `--frameworks`, `--tests`, and `--benchmarks`.
9. Removed `jit-diff install --last_successful` option because it didn't work.
10. Removed the two old sample json files (that didn't work), and replace them with
    a sample config.json file that does.


BruceForstall:
@russellhadley @JosephTremoulet @dotnet/jit-contrib PTAL

doc/diffs.md
```diff
@@ -68,53 +166,184 @@ The "jit-diff list" command has this help message:
...
+* Which base and diff JIT or crossgen to use.
+* Which crossgen to use (if base and diff are specified as JIT, not crossgen).
+* Which assemblies to generate dasm for.
+* Where to put the generated dasm.
+
+These can all be specified explicitly. For example:
+
 ```
-    c:\gh\coreclr> jit-diff diff -o c:\diffs -t 5 -a -f --core_root c:\gh\coreclr\bin\tests\Windows_NT.x86.checked\Tests\Core_Root -b e:\gh\coreclr2\bin\Product\Windows_NT.x86.checked -d c:\gh\coreclr\bin\Product\Windows_NT.x86.checked
+    c:\gh\coreclr> jit-diff diff --output c:\diffs --corelib --core_root c:\gh\coreclr\bin\tests\Windows_NT.x64.checked\Tests\Core_Root --base e:\gh\coreclr2\bin\Product\Windows_NT.x64.checked --diff c:\gh\coreclr\bin\Product\Windows_NT.x64.checked
```
JosephTremoulet: Should use `release` for `core_root`

doc/diffs.md
```diff
@@ -68,53 +166,184 @@ The "jit-diff list" command has this help message:
...
+most of the arguments using computed defaults. The above diff can be accomplished using simply:
+```
+    c:\gh\coreclr> jit-diff diff --base_root e:\gh\coreclr2
+```
+
+The defaults are:
+* If jit-diff is invoked with the current directory within the dotnet/coreclr repo, then the root of
+  this repo serves to find the diff compiler and `CORE_ROOT` directory. (Note that we have no reasonable
+  default for determining what the baseline toolset or repo is. This is specified with the `--base_root` argument.)
+* The default output directory is `<repo_root>\diffs` (in this case, c:\gh\coreclr\diffs).
```
JosephTremoulet: Not a fan of spewing output into the source tree.  Does coreclr's `.gitignore` ignore this?

doc/diffs.md
```diff
@@ -68,53 +166,184 @@ The "jit-diff list" command has this help message:
...
+
+As seen above, specifying all required information can be quite verbose. jit-diff can automatically determine
+most of the arguments using computed defaults. The above diff can be accomplished using simply:
+```
+    c:\gh\coreclr> jit-diff diff --base_root e:\gh\coreclr2
+```
+
+The defaults are:
+* If jit-diff is invoked with the current directory within the dotnet/coreclr repo, then the root of
+  this repo serves to find the diff compiler and `CORE_ROOT` directory. (Note that we have no reasonable
```
JosephTremoulet: So if I want to generate baselines only, I have to run this tool from outside the repo directory structure?  That's counter-intuitive.

doc/diffs.md
```diff
@@ -68,53 +166,184 @@ The "jit-diff list" command has this help message:
...
+  default for determining what the baseline toolset or repo is. This is specified with the `--base_root` argument.)
+* The default output directory is `<repo_root>\diffs` (in this case, c:\gh\coreclr\diffs).
+* The default diff architecture is x64. If this isn't found, x86 is tried.
+* The default diff build flavor is checked. If this isn't found, debug is tried.
+* By default, diffs are done using System.Private.CoreLib.dll. (That is, `--corelib` is the default.)
```
JosephTremoulet: Jit build flavor and input build flavor should be separate.  Release should be the default for inputs.

doc/diffs.md
```diff
@@ -68,53 +166,184 @@ The "jit-diff list" command has this help message:
...
+
+To instead do diffs over the framework assemblies (not just System.Private.CoreLib.dll), using an x86
+debug build, run:
+```
+    c:\gh\coreclr> jit-diff diff --frameworks --arch x86 --build debug --base_root e:\gh\coreclr2
+```
+
+To simplify this more, create a dotnet/coreclr repo clone that you will always use for baselines.
+Create a configuration file that specifies your baseline root directory, specifying a `base_root`
+default. See the document [configuring defaults](config.md) for details.
```
JosephTremoulet: Really?  Why not use cijobs to grab a baseline jit?  Or checkout baseline and generate base then checkout diff and generate diff?  I'd think those should all be mentioned as options, and the tool seems to be explicitly making the third one (my favorite) difficult.

doc/diffs.md
```diff
@@ -68,53 +166,184 @@ The "jit-diff list" command has this help message:
...
+* `--build`
+
+## Using tags
+
+jit-diff takes an optional '--tag' command-line argument.  This tag can be used to label different
+directories of `*.dasm` in the output directory so multiple runs can be done.
+This supports a scenario like the following:
+
+* Build base CoreCLR
+* Produce baseline diffs by invoking the tool with '--base'
```
JosephTremoulet: This has to be done from outside the coreclr tree, or I'm misreading?

doc/diffs.md
```diff
@@ -68,53 +166,184 @@ The "jit-diff list" command has this help message:
...
+to identify files missing from the base dataset versus the diff dataset.
+
+For the simplest case just point the tool at a base and diff directory produce by jit-diff and it
+will summarize code size differences across the whole diff. This is what the jit-diff command lines in
+the previous section do.
+
+On a significant set of diffs it will produce output like the following:
+
+```
+$ jit-analyze --base ~/Work/dotnet/output/base --diff ~/Work/dotnet/output/diff
```
JosephTremoulet: Example should use `--recursive`

doc/formatting.md
```diff
@@ -42,19 +43,6 @@ Sample help command line
...
 
-## Dependencies
-
-* clang-format - The jit-format tool calls clang-format to run whitespace based formatting changes.
-* clang-tidy - The jit-format tool calls clang-tidy to run more advanced formatting changes.
-
-Running `bootstrap.cmd` or `bootstrap.sh` will download the correct version of clang-format and clang-tidy
-for your platform, if the correct version is not already on your path.
-
-We require version 3.8.0 for jit-format. If you wish to install it yourself,
```
JosephTremoulet: Why did you get rid of this?  I'd rather not use bootstrap and re-download clang-tidy whenever I want to build.

doc/config.md
```diff
@@ -0,0 +1,123 @@
...
+      "verbose": "true",
+      "fix": "true"
+    }
+  },
+  "asmdiff": {
+    "default": {
+      "base": "checked_osx-1526",
+      "diff": "/Users/russellhadley/Work/dotnet/coreclr/bin/Product/OSX.x64.Checked",
+      "analyze": "true",
+      "frameworksonly": "true",
```
JosephTremoulet: Likewise for `analyze`

src/jit-diff/jit-diff.cs
```diff
@@ -159,25 +184,240 @@ private void SetRID()
...
+
+                    Console.WriteLine("Using --output={0}", _outputPath);
+                }
+
+                bool needTestTree = Benchmarks || DoTestTree;
+
+                if ((_platformPath == null) || (_basePath == null) || (_diffPath == null) || ((_testPath == null) && needTestTree))
+                {
+                    // Try all combinations of build and architecture to find one that has both a Core_Root
+                    // and, if necessary, base and diff builds and tests.
```
JosephTremoulet: `core_root` and `test_root` should only ever default to `release`.  If the person running the tool hasn't built the release tree, better to tell them they need to than have them run diffs on a different build flavor.

src/jit-diff/jit-diff.cs
```diff
@@ -159,25 +184,240 @@ private void SetRID()
...
                     }
-                    break;
-                    default:
+
+                    List<string> buildList;
+                    if (_build == null)
                     {
-                        Console.WriteLine("No platform mapping!");
-                        _moniker = "bogus";
+                        buildList = new List<string> { "checked", "debug" };
```
JosephTremoulet: Defaulting build flavors for any of these seems dicey.  I for one almost always have out-of-date trees built of one flavor or another.  I'd go so far as to say we should print out what we're picking and prompt for user input before continuing (maybe with an explicit `--quiet` flag to opt out).

src/jit-diff/jit-diff.cs
```diff
@@ -159,25 +184,240 @@ private void SetRID()
...
+                    foreach (var build in buildList)
+                    {
+                        foreach (var arch in archList)
+                        {
+                            var buildDirName = GetBuildOS() + "." + arch + "." + build;
+                            string tryPlatformPath = null, tryBasePath = null, tryDiffPath = null, tryTestPath = null;
+
+                            if ((_platformPath == null) && (_diffRoot != null))
+                            {
+                                tryPlatformPath = Path.Combine(_diffRoot, "bin", "tests", buildDirName, "Tests", "Core_Root");
```
JosephTremoulet: On Linux, I've always pointed `--core_root` to the overlay directory I created.  I think you've mentioned in the docs above that running `runtests.sh GenerateLayoutOnly` will populate this `Core_Root` with the right set of stuff that this will work on Linux, but wanted to make sure I'm getting that right -- is that what you're saying/expecting, that this formula works on all supported platforms assuming that the dev has done a `GenerateLayoutOnly` runtests?

src/jit-diff/jit-diff.cs
```diff
@@ -891,27 +1230,38 @@ private void StartDasmWorkSingle(List<string> args, string assemblyPath)
...
+                string outputPath = Path.Combine(m_config.OutputPath, tagBaseDiff, assemblyInfo.OutputPath);
+                args.Add("--output");
+                args.Add(outputPath);
+
+                args.Add(assemblyInfo.Path);
+
+                StartDasmWork(args);
+            }
+
             // Returns a count of the number of failures.
```
JosephTremoulet: Stale comment

src/jit-format/jit-format.cs
```diff
@@ -115,9 +118,57 @@ private void SetPlatform()
                 }
             }
 
+            private string GetRepoRoot()
```
JosephTremoulet: I wonder if we shouldn't have a jitutils.utility.dll...

doc/diffs.md
mikedn: This repo clone thing seems like an unnecessary complication. Personally I just do a build from the master branch (which is normally clean) and copy some of the output binaries somewhere else (e.g. Windows_NT.x64.Checked_base).

doc/config.md
```diff
@@ -0,0 +1,123 @@
...
+	checked_osx-1439
+	checked_osx-1442
+	checked_osx-1443
+	checked_osx-1526
+```
+
+### Installing new tools
+
+The jit-diff command `install` will download and install a new tool to the default location
+and update the config.json so it can be found.
```
BruceForstall: No; `install` still exists. Honestly, though, I've never used it (successfully).

doc/diffs.md
```diff
@@ -1,65 +1,163 @@
...
+* Of the test tree, only the benchmarks. Use `--benchmarks`.
+
+`--corelib` is the default.
+
+Only one of `--corelib` or `--frameworks` can be specified.
+
+Only one of `--tests` or `--benchmarks` can be specified.
+
+If `--tests` or `--benchmarks` is specified, you must also specify `--test_root` so the tool knows
+where to find the test tree you wish to use.
```
BruceForstall: It does (that was a late change). I'll update the wording.

doc/diffs.md
BruceForstall: Interesting point. What would your suggested usage be? Another flag, e.g. `--baselineonly`? `--nodiff`?

My mental model (and normal mode of operation) is that you are working in the repo, making changes, and want to generate diffs using those changes. I never just generate baselines. Maybe I don't iterate on just diffs frequently. I think you can still generate base/diff the first time, then just diffs afterwards. Although if you specify `--base_root`, say in your config.json, the defaults will cause you to always generate baselines.

doc/diffs.md
BruceForstall: We could change `.gitignore`. Or, you could pass `--output` or set it in a config.json.

Do you have an alternative suggestion for a default?

doc/diffs.md
JosephTremoulet: I'd be inclined toward something like `--diffRootCwd` to opt in to defaulting diff from current directory, rather than keying off whether you're in the tree.

doc/diffs.md
JosephTremoulet: My inclination is to make `--output` a required parameter.  Failing that, I don't see what to do other than put `diffs` in `.gitignore` or maybe default to `obj/diffs` or `bin/diffs` if one of those is appropriate.

doc/diffs.md
BruceForstall: That's quite complicated. Why wouldn't you just use the same build flavor? Or, specifically, why shouldn't the default use the same build flavor? I almost never build and use release, for example.

doc/config.md
russellhadley: I've used install successfully - this is a time saver for me so I'm keen to keep it.

doc/diffs.md
BruceForstall: I would prefer if the tool automatically found and downloaded an appropriate baseline. As that seems difficult (or impossible), it seems the user needs to specify it. Another option: the tool automatically clones and builds a baseline that is correct w.r.t. your source control point. Certainly too expensive. Another: the tool is given complete control over a clone and sync's and incrementally builds the baseline.

Yes, Joe: the options you mention could be given as examples, and made to work easily. There some tension between simplicity of each type of option.

> This repo clone thing seems like an unnecessary complication. Personally I just do a build from the master branch (which is normally clean) and copy some of the output binaries somewhere else (e.g. Windows_NT.x64.Checked_base).

I find that to be too error prone. Also, if I forget to build/save master, I need to (a) change branch, (b) build, (c) change back, (d) rebuild again. The (d) step is annoying, especially if you change VM/JIT common header files. I find it easier to just keep a baseline clone around left untouched, and sync'ed at the same time as my diff branch.

doc/diffs.md
russellhadley: Just to add a wrinkle here, I typically run from outside the repo structure and provide a standalone output dir with tools in it.

doc/diffs.md
BruceForstall: Yeah, I guess there's a usage problem here where the defaults will kick in and find the diff compiler. Maybe there should be a way to prevent this. Maybe if you specify --base or --diff directly, the defaults don't kick in.

doc/diffs.md
russellhadley: +1 for separate.

doc/formatting.md
BruceForstall: bootstrap doesn't re-download unless it can't find clang-format/clang-tidy.

I just moved this text elsewhere.

src/jit-format/jit-format.cs
BruceForstall: I'd prefer to just have a shared utils.cs file statically linked in. But yes, there is too much code duplication now.

src/jit-diff/jit-diff.cs
BruceForstall: I'm not sure I agree. I never use release. And the tool does tell you when it uses a default, what it is.

src/jit-diff/jit-diff.cs
BruceForstall: That's my expectation.

doc/diffs.md
russellhadley: I think I'm still leery about putting non config file defaults baked into the tool.  Just in the 4 people that have commented we've seen enough variation to make this a hard nut to crack.  I'm worried that we're inviting forking the tools to support different peoples defaults.

doc/diffs.md
JosephTremoulet: The reason to diff release frameworks dlls is so that the diffs we're looking at are the impact our compiler changes are having on the binaries we ship as part of the product.
With benchmarks, making sure you're diffing with optimizations enabled is important -- you don't need release/checked for that after Andy's recent changes, but release still seems safest, especially if we ever get to more jit-like codegen and pull in e.g. inlines from frameworks dlls.

doc/diffs.md
JosephTremoulet: > I would prefer if the tool automatically found and downloaded an appropriate baseline. As that seems difficult (or impossible)

I'm not following why cijobs couldn't be expanded to make that doable.  Seems like the main things lacking are:
 - cijobs having a way to find a job that matches a given commit hash, instead of having to specify by jenkins run ID
 - jit-diff having a way to figure out what baseline you want (might be as simple as `git merge-base HEAD coreclr/master`)
 - jit-diff knowing which jobs build which compilers

>I find that to be too error prone... I find it easier to just keep a baseline clone around left untouched, and sync'ed at the same time as my diff branch.

Sure, you do you.  I find it too error prone *not* to do it that way.  The main point is that the tool needs to support all our workflows.

doc/formatting.md
JosephTremoulet: > bootstrap doesn't re-download unless it can't find clang-format/clang-tidy

What does it do when we update the clang-format/clang-tidy in Azure to a newer version?

src/jit-diff/jit-diff.cs
JosephTremoulet: >  And the tool does tell you when it uses a default, what it is.

Sure, but that will be buried by the output spew, hence the desire for a prompt before spewing.

doc/formatting.md
BruceForstall: Nothing.

If that ever happens, you'd need to re-download them. 2 minute step: "git clean -dxf; bootstrap.cmd"

doc/config.md
BruceForstall: Nobody has suggested removing it.

doc/diffs.md
BruceForstall: How about: pass "--base" if you want baselines, "--diff" if you want diffs, or pass both. If you don't pass any argument (assuming the command-line parser handles this), then use defaults. If you pass the path, use that. This unfortunately means the minimum arguments are increased to:

```
jit-diff diff --diff
```

Or, if you want baselines too:
```
jit-diff diff --diff --base --base_root c:\gh\coreclr_baseline
```

All the previous usage models still exist (passing everything explicitly). It's only that defaults are computed when you don't specify them all.

doc/diffs.md
BruceForstall: I like the idea of (by default) putting them in bin/diffs; that directory is already .gitignore'd.

BruceForstall: I don't want to have *any* required parameters if possible.

doc/diffs.md
BruceForstall: You can always just pass everything explicitly, as today. Is that not sufficient?

I can also see having a different set of build flavor defaults for core_root/test_root and the JIT directories, and not trying to get them all to match when computing defaults. I'm leery of adding additional `--build` arguments, though: it seems like you should just pass the path explicitly if you don't like the default.

It seems like you are suggesting:
`--core_root` and `--test_root` should default to release, then fall back to checked or debug. (Or not fall back at all? I can't agree with that option, since my goal is to have full defaults for everything.)
`--diff` and `--base` should default to debug, then fall back to checked (or vice-versa, as I currently have it?)

I normally build a checked JIT before doing diffs, as it's faster.

doc/diffs.md
JosephTremoulet: `--base` and `--diff` currently take arguments, they wouldn't in your latest suggestion.  So then we could have new zero-argument flags `--base` and `--diff`, and rename the old ones to something like `--base_path` and `--diff_path`?  Then we could also have `--diff_path` and `--diff_root` imply `--diff`, and likewise `--base_path` and `--base_root` imply `--base`... I think I'd be ok with that, if the help text explained the options reasonably well.

doc/diffs.md
JosephTremoulet: > You can always just pass everything explicitly, as today. Is that not sufficient?

That's sufficient to generate diffs for whatever inputs you want.  I have an additional "pit of success" goal that someone following the path of least resistance to generate diffs will generate diffs over the most meaningful inputs, which are definitely release frameworks and likely release benchmarks.

>`--core_root` and `--test_root` should default to release, then fall back to checked or debug. (Or not fall back at all? I can't agree with that option, since my goal is to have full defaults for everything.)

Default to release, not fall back at all.  I don't understand why it's important to you that we find something to run diffs over if the salient inputs aren't built on the person's machine; we may have to agree to disagree.

> `--diff` and `--base` should default to debug, then fall back to checked (or vice-versa, as I currently have it?

vice-versa, as you currently have it, though it would be nice if there were *something* to avoid using stale bits, like checking that the one we pick isn't older than the one we skip or ...

> I normally build a checked JIT before doing diffs, as it's faster

Me too, with `skiptests`.  And a release test tree, since it's what I think we should all be running diffs over.

doc/diffs.md
BruceForstall: No, I was suggesting that `--base` and `--diff` would take arguments (a path) as today, which would do what they do today, but the argument would be optional. If you didn't pass a path, it would try to use a default. It wouldn't depend on `--diff_root` being set, for example, because you might want to set and leave that set in the config.json file.

So, you would have:
```
jit-diff diff
  -- illegal: no --base or --diff specified, so don't know what to diff
jit-diff diff --diff
  -- do a "diff" generation. Needs to use a default to find the diff JIT. It either uses the current
      directory or `--diff_root` in config.json to find the repo root, then uses default arch/build
      (optionally overridden by `--arch`, `--build` to find the JIT compiler.)
jit-diff diff --base
   -- do a "baseline" asm generation.
jit-diff diff --diff --base
   -- do both baseline and diff generation
jit-diff diff --base e:\gh\coreclr2\bin\Product\Windows_NT.x64.checked
   -- do just a baseline generation, using my given directory
jit-diff diff -diff --base e:\gh\coreclr2\bin\Product\Windows_NT.x64.checked
   -- do both diff and baseline generation, using my given directory for the baseline, and
      discovered defaults for the diff JIT.
```

Presumably only if both `--base` and `--diff` are specified could you invoke jit-analyze. Unless there are reasonable base/diff defaults that could be used. E.g., if you (a) generate baselines, then (b) sequentially generate a series of diff compiles -- could jit-analyze automatically (or non-automatically) find the baseline originally created in (a) for each run of (b)?

JosephTremoulet: I didn't realize that named flags could have optional arguments.  Your list of possible inputs (if I add some examples with an explicit path after `--diff`) looks like the list of invocations I was trying to get at in my last comment (thanks for clarifying), so yes that sounds workable to me.

I'd rather just invoke jit-analyze manually on whichever pairs I'm interested in that have ways to point it at or have it hunt down baselines.

BruceForstall:
@JosephTremoulet @russellhadley I've updated jit-diff with new argument rules, as discussed in the PR comments. I've updated all the help, as well as adding a bunch of examples to the tool usage message. E.g., you must now specify `--base` to generate baselines, `--diff` to generate diffs. If these aren't given arguments (full paths), defaults are found. Also, `--core_run`, `--crossgen`, and `--test_root` default to release builds, if available.

PTAL.

Also, I would appreciate it if you also pulled this down and tried it on your scenarios to ensure that it works for those, which are different from mine.

cc @dotnet/jit-contrib

russellhadley:
I just noticed the -lastSuccessful flag being removed.  What happened? It broke once before when the Jenkins REST API output changed. Was it that again?  (I find it convenient to grab the last successful build as a baseline and then branch from there for working since I'm in a good state then by construction)

russellhadley:
I figured out the mystery with why install doesn't work.  The dotnet moniker for Windows changed and no longer matches the directory name I was looking for in what Jenkins produces for the job.  There's some missing defensive code that needs to get added.  This can be fixed by using the new GetBuildOS method, but this highlights that there's duplication between the SetRID logic and GetBuildOS that should be reconciled.

russellhadley:
I've fixed the install command and installed my base tool set and have everything set in the config file but I'm having a lot of trouble running the baseline set:
```
[D:\dotnet]
 10:12:25 > jit-diff diff --base --core_root D:\dotnet\coreclr\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root
Default diff path D:\dotnet\coreclin\Product\Windows_NT.x64.Checked not found! Investigate config file entry.
Environment variable JIT_UTILS_ROOT found - configuration loaded.
Error: didn't find --crossgen default
Beginning diff of System.Private.CoreLib.dll, framework assemblies
crossgen not found at D:\dotnet\output\tools\checked_windows_nt-1337\Product\Windows_NT.x64.Checked\crossgen.exe

Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at ManagedCodeGen.jitdiff.DiffTool.StartDasmWorkOne(List`1 commandArgs, String tagBaseDiff, String clrPath, AssemblyI
nfo assemblyInfo)
   at ManagedCodeGen.jitdiff.DiffTool.StartDasmWorkBaseDiff(List`1 commandArgs, List`1 assemblyWorkList)
   at ManagedCodeGen.jitdiff.DiffTool.RunDasmTool(List`1 commandArgs, List`1 assemblyWorkList)
   at ManagedCodeGen.jitdiff.DiffTool.DiffCommand(Config config)
   at ManagedCodeGen.jitdiff.Main(String[] args)
[D:\dotnet]
 10:13:28 > ls D:\dotnet\output\tools\checked_windows_nt-1337\Product\Windows_NT.x64.Checked\crossgen.exe


    Directory: D:\dotnet\output\tools\checked_windows_nt-1337\Product\Windows_NT.x64.Checked


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         4/4/2017   7:39 AM        4547584 crossgen.exe

```

russellhadley:
More details
I'm running out side of my enlistment directory (I don't like keeping my output in my source repos)
(I PR'd your private fork with the fix for install so you should be able to repro this)

And here's my config file:
```
{
  "asmdiff": {
    "default": {
      "base" : "checked_windows_nt-1337",
      "diff" : "D:\\dotnet\\coreclr\bin\\Product\\Windows_NT.x64.Checked",
      "frameworks": "true",
      "output": "d:\\dotnet\\output"
    },
    "tools": [
      {
        "tag": "checked_windows_nt-1337",
        "path": "D:\\dotnet\\output\\tools\\checked_windows_nt-1337\\Product\\Windows_NT.x64.Checked"
      }
    ]
  },
  "format": {
    "default": {
      "arch": "x64",
      "build": "Checked",
      "os": "Windows_NT",
      "coreclr": "C:\\dotnet\\coreclr",
      "verbose": "true",
      "fix": "true",
      "untidy": "false"
    }
  }
}
```

src/jit-diff/jit-diff.cs
```diff
@@ -159,23 +224,340 @@ private void SetRID()
...
+                            {
+                                Console.WriteLine();
+                                Console.WriteLine("Warning: it is best practice to use a release build for --core_root, --crossgen, and --test_root.");
+                                Console.WriteLine();
+                            }
+
+                            break;
+                        }
+
+                        if (!needCoreRoot && !needCrossgen && !needBasePath && !needDiffPath && !needTestTree)
```
JosephTremoulet: If you can find an appropriate basePath and diffPath on x64 but can't find one of coreRoot/crossgen/testTree on x64, or similarly if you can find coreRoot and crossgen and testTree on x64 but can't find one of basePath/diffPath, and you can find all 5 on x86, it looks like this logic would lock in the set it finds on x64 and then just find the others on x86 and then try to run using defaults with mismatched architectures.

src/jit-diff/jit-diff.cs
```diff
@@ -234,93 +616,178 @@ private void ExpandToolTags()
...
 
-                if (_outputPath == null)
+                if (_benchmarks && _tests)
+                {
+                    DisplayErrorMessage("Specify only one of --benchmarks or --tests");
+                }
+
+                if (_benchmarks && (_testPath == null))
+                {
+                    DisplayErrorMessage("--benchmarks requires specifying --test_root as well");
```
JosephTremoulet: --test_root or --diff_root or running in the coreclr tree

src/jit-diff/jit-diff.cs
```diff
@@ -234,93 +616,178 @@ private void ExpandToolTags()
...
+                }
+
+                if (_tests && (_testPath == null))
                 {
-                    _syntaxResult.ReportError("Specify --output <path>");
+                    DisplayErrorMessage("--tests requires specifying --test_root as well");
```
JosephTremoulet: Ditto re: default test_root

src/jit-diff/jit-diff.cs
```diff
@@ -234,93 +616,178 @@ private void ExpandToolTags()
...
                 }
 
-                if (!Directory.Exists(_outputPath))
+                if (!_corelib && !_frameworks && !_benchmarks && !_tests)
                 {
-                    _syntaxResult.ReportError("Can't find --output path.");
+                    // Setting --corelib as the default
```
JosephTremoulet: Maybe move this to `SetDefaults()`?

JosephTremoulet:
Forgot to mention:  I did pull down and try out the latest changes, and yes they do work fine in my typical workflow.

BruceForstall:
I've updated the changes with to reflect code review feedback, as well as:

- Fixed many more crashing issues.
- Fixed `jit-diff install --last_successful` so it works.
- Added `jit-diff uninstall`
- Refactored the code: created a shared src/util/util.cs file and put some common functions there. Also split jit-diff.cs: added diff.cs, install.cs, uninstall.cs to contain code specific to those commands.

cc @dotnet/jit-contrib

BruceForstall:
About the only thing I didn't address were Joe's two top-level concerns: (1) worry about errors getting lost in the output spew. I think it would be easy to capture all the error/warning messages and replay them at the end. Future work. (2) worry about the strategy of iterating arch/build to find defaults.

JosephTremoulet:
- The change to cijobs copy --output behavior complicates using it (with -c) in scripts; it used to be the case that the calling script would know exactly where the output was going to go, but now it would have to figure out the job number.  Not sure what to do about it:
    - We could have separate --output (old behavior) and --outputRoot (new behavior) parameters
    - We could have the directory tag match the inputs (so use the commit hash when called with -c)
    - You or I could just deal with having it work the way we don't like (I know that for me that would involve having calling scripts search the output directory or parse cijobs' stdout; I don't know what it would involve for you).
 - I often have to run cijobs copy a few times before it succeeds, due to network flakiness.  If it's now an error to specify an existing --output directory, I'd have to delete the directory between retries.  So I'd rather that be a warning, or an error suppressible with a flag like --force, or have cijobs copy detect failure and delete the directory if the copy fails.

Aside from that, LGTM, thanks.

BruceForstall:
I thought about the cijobs output path, but hoped the change wouldn't cause problems. Of course, I was wrong :-)  I actually thought it might have been a bug, since the "tag" (final directory name component) was already being constructed, but never used. I'll probably need to add `--outputRoot` since jit-diff requires that behavior.

I can make the cijobs copy existing output file check a warning, no problem. When I had that happen, cijobs crashed; I think in the unzip phase. Maybe I can look more carefully at what caused the crash.

BruceForstall:
@dotnet/jit-contrib @dotnet/arm64-contrib @dotnet/arm32-contrib I am planning to merge this change soon. Note that the command-line for jit-diff changes somewhat drastically with this change (although mostly in a backwards-compatible way). If you have any questions, let me know.

PR #104, opened 2017-04-11
Suppress cijobs copy timeout

JosephTremoulet:
The default of 100 seconds is often too short.  Set it to infinite, so
user or automation harness can control how long to wait before killing.


JosephTremoulet:
@dotnet/jit-contrib PTAL

russellhadley:
LGTM

PR #108, opened 2017-04-18
Use crossgen.exe from core_root

JosephTremoulet:
When --crossgen is not explicitly passed, pull it from --core_root rather
than --base or --diff; --crossgen and --core_root need to agree in build
flavor, but --base and --diff do not need to agree with them, so the
defaults of release --core_root and checked --base/--diff result in an
error otherwise.  This also helps ensure that crossgen.exe runs on the
corelib right next to it.


JosephTremoulet:
@dotnet/jit-contrib

PR #109, opened 2017-04-25
Separate base/diff/exception dasm error counts

JosephTremoulet:
This gives a quick indication of whether diffs failed more with one jit
than the other.

### DIFF
--- a/src/jit-diff/diff.cs
+++ b/src/jit-diff/diff.cs
@@ -19,8 +19,24 @@ using System.Threading.Tasks;
 
 namespace ManagedCodeGen
 {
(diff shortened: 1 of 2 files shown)


JosephTremoulet:
Output looks like so:

```
...
assembly name: xunit.execution.desktop.dll
assembly name: xunit.execution.desktop.dll
Dasm commands returned 857 base failures, 857 diff failures.
Analyze command: ...
...
135 total methods with size differences (132 improved, 3 regressed).

Warning: Failures detected generating asm: 857 base, 857 diff
```

/cc @dotnet/jit-contrib

russellhadley:
LGTM

PR #114, opened 2017-07-14
Add scripting to download xml config files.

jashook:
This will allow the coreclr xml files to be diffed. Note that
there is some timing that requires the ci generate job to have
run.

@RussKeldorph @mmitche ptal

The only issue this seems to have is every Http request requires an OAuth to Github which after 2k+ authenticated get requests I seem to be blocked by Github to authenticate further.

This keeps the script from finishing and requires you to wait about 30 minutes before continuing.

### DIFF
--- a/src/coreclr_config_download/README.md
+++ b/src/coreclr_config_download/README.md
@@ -1,0 +1,60 @@
+**CoreClr Config Download**
+
+Download all of the config.xml files for each coreclr job. This will will download into two different folders <output_location>/base and <output_location/diff>.
+
+You can then diff the two folders with whatever your favorite diff tool is. An example on unix would be:
+
+>$diff -rq <output_location>/base <output_location>/diff
+
+**Requirements**
+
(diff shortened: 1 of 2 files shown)


jashook:
Also ptal @adiaaida 

/cc @russellhadley

RussKeldorph:
@pgavlin 

I haven't looked in detail yet, but why is github required for this?  Shouldn't this all be queryable via unauth GET requests to Jenkins?  I'm sure I'm missing something obvious.

jashook:
Jenkins considers the config.xml request to require authentication. If there is a way around this that I have not seen then I would love to know. It currently is quite a large downside.

RussKeldorph:
Oh yeah, that's right.  You have to be auth'd to see the configure link.  Sorry I forgot about that.

pgavlin:
What is in the config.xml files, and why do we need to be able to diff them?

jashook:
The config files define each job's execution and all other properties. A no diff would prove that an old job will not be modified.

pgavlin:
Okay. Seems fine to me, then.

PR #115, opened 2017-07-24
Support netcoreapp2.0

sbomer:
Adds a `<TargetFrameworks>` element to each project that includes
netcoreapp2.0 as a target. This is useful because it makes it possible
to build and run jitutils with a 2.0 install of the cli, which doesn't
come with a 1.0 runtime.

For now, the default target is still netcoreapp1.0, to prevent
breaking ci jobs that run jitutils with a 1.0 runtime.

### DIFF
--- a/src/cijobs/cijobs.csproj
+++ b/src/cijobs/cijobs.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFrameworks>netcoreapp1.0;netcoreapp2.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/jit-analyze/jit-analyze.csproj
+++ b/src/jit-analyze/jit-analyze.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFrameworks>netcoreapp1.0;netcoreapp2.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/jit-dasm/jit-dasm.csproj
+++ b/src/jit-dasm/jit-dasm.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFrameworks>netcoreapp1.0;netcoreapp2.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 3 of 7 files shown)


BruceForstall:
cc @JosephTremoulet @russellhadley @dotnet/jit-contrib

sbomer:
This won't introduce any changes to the 1.0 workflow as far as I'm aware.

JosephTremoulet:
LGTM, thanks.

> For now, the default target is still netcoreapp1.0, to prevent breaking ci jobs that run jitutils with a 1.0 runtime.

Should there be a tracking issue to come back when the ci jobs change?

> makes it possible to build and run jitutils with a 2.0 install of the cli, which doesn't come with a 1.0 runtime.

Does `dotnet restore ; dotnet publish` "just work" on a 2.0 install of the cli (with this change), or do you have to explicitly request `netcoreapp2.0` in one of those steps (or somewhere else)?

sbomer:
> Should there be a tracking issue to come back when the ci jobs change?

The only jobs I'm aware of that would be impacted are the jit-format runs, and I don't know whether there are plans to change these. @adiaaida 

I think it's mostly a matter of making 2.0 the default for jitutils, depending on whether folks are ok with installing a new cli, and then updating the jobs to match. So it might be worth making an issue to track making 2.0 the default.

> Does dotnet restore ; dotnet publish "just work" on a 2.0 install of the cli (with this change), or do you have to explicitly request netcoreapp2.0 in one of those steps (or somewhere else)?

You have to explicitly request netcoreapp2.0 during publish: `dotnet publish -f netcoreapp2.0`

sbomer:
I went ahead and opened a tracking issue: https://github.com/dotnet/jitutils/issues/116
Thanks all!

JosephTremoulet:
> You have to explicitly request netcoreapp2.0 during publish: `dotnet publish -f netcoreapp2.0`

Presumably we should add that to the documentation, then.  I just took a look (figured PRing the doc change would be just as easy as commenting that we should make it), though, and noticed that we're directing people toward `bootstrap.cmd/sh` and `build.cmd/sh`, neither of which were updated in this PR -- so presumably the build steps in the directions don't work on a 2.0 install of dotnet cli?  Are you working on a change to update them, should we have an issue?

adiaaida:
> You have to explicitly request netcoreapp2.0 during publish: dotnet publish -f netcoreapp2.0

Doesn't that just mean that it's going to be published in the netcoreapp2.0 folder? I'm not sure that does anything other than tell publish what folder to put the results in.

sbomer:
I believe it's equivalent to `/p:TargetFramework=netcoreapp2.0`, which tells MSBuild to set the TargetFramework property as if it had been set in the project file, and also changes the name of the publish folder.

sbomer:
> so presumably the build steps in the directions don't work on a 2.0 install of dotnet cli? Are you working on a change to update them, should we have an issue?

Well, the build steps work, but the 2.0 cli can't run the output because the projects target 1.0. The issue I filed tracks making 2.0 the default in the project files and build scripts: https://github.com/dotnet/jitutils/issues/116. I think `bootstrap.cmd/sh` and `build.cmd/sh` should be updated together with the docs when we decide to make the change.

Issue #107, opened 2017-04-17
Add defaults for cijobs


BruceForstall:
Add `--arch`, `--os`, and `--build` arguments to cijobs. They would automatically map to Jenkins job ids. They could be defaulted using the config.json file. Then, you could do:
```
cijobs list --arch x86 // only show the x86 jobs
cijobs list --os ubuntu // only show Ubuntu jobs
cijobs list --arch x86 --os ubuntu // only show x86 ununtu jobs
cijobs copy -l --arch x64 --os windows --build checked // copy the latest x64/Windows/checked build
cijobs copy -l // same as above, if the standard defaults or config.json file specify x64/Windows/checked
```

Also, `jit-diff install` should match `cijobs copy` functionality.

PR #118, opened 2017-08-15
Add --server parameter to cijobs to access ci2 and ci3

sbomer:
Currently cijobs isn't able to retrieve jobs from http://ci2.dot.net/ or http://ci3.dot.net/.


sbomer:
I'll update the documentation to mention this flag, thanks!

Currently one just needs to be familiar enough with the ci jobs. Right now the cijobs tool is essentially just a front-end for jenkins - it knows nothing about the coreclr jobs, aside from the default values for the jobName, server, etc.

We could have logic to guess jenkins jobs essentially mapping job configurations to jenkins jobs similar to the suggestion in https://github.com/dotnet/jitutils/issues/107. This would involve giving cijobs some knowledge of how the jenkins jobs for coreclr are defined, resulting in a new dependency that would require updates in cijobs to match changes or additions to job definitions in [netci.groovy](https://github.com/dotnet/coreclr/blob/0bd583ccecc46d106ccb6af9f2f1e75f515460f8/netci.groovy).

sbomer:
Also, I think most of the jobs people will be interested in run on ci.dot.net today. I'm introducing this flag just to make it easier for me to experiment with new jobs on ci3.dot.net.

Issue #116, opened 2017-07-24
Target netcoreapp2.0 by default


sbomer:
At some point we should update jitutils to default to targeting netcoreapp2.0, so that the build scripts "just work" with a .NET Core 2.0 cli install.

This will be a "breaking change" to the build steps, requiring an update to the 2.0 cli, and will probably require minor changes to the jit-format jobs.

See https://github.com/dotnet/jitutils/pull/115 for some discussion.

sandreenko:
Thanks, have checked that it works for me. It only works from "Developer Command Promt", when the standard cmd shows: "There was an error running CMake to generate compile_commands.json. Please do a full build to generate a build log."

BruceForstall:
@sbomer This can be closed now, right?

BruceForstall:
I think this is no longer an issue.

PR #121, opened 2017-10-24
Target netcoreapp2.0 by default

sbomer:
Since a few people have hit problems when trying to use jitutils with a 2.0 dotnet install, I think it's time to address https://github.com/dotnet/jitutils/issues/116.

With this change, you'll need to install the 2.0 dotnet SDK to build jitutils.

@adiaaida, I think at some point the jit-format jobs were building jitutils using some specific version of the SDK. This might require the format jobs to be updated if they still do that.

/cc @sandreenko

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -15,6 +15,7 @@ set fxInstallDir=%scriptDir%fx
 set buildType=Release
 set publish=false
 set fx=false
+set tfm=netcoreapp2.0
 
 for /f "usebackq tokens=1,2" %%a in (`dotnet --info`) do (
     if "%%a"=="RID:" set platform=%%b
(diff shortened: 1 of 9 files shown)


sbomer:
Fixed, thanks!

build.cmd
```diff
@@ -77,11 +83,12 @@ exit /b 0
...
 :usage
 echo.
-echo  build.cmd [-b ^<BUILD TYPE^>] [-f] [-h] [-p]
+echo  build.cmd [-b ^<BUILD TYPE^>] [-f] [-h] [-p] [-t ^<TARGET^>]
 echo.
 echo      -b ^<BUILD TYPE^> : Build type, can be Debug or Release.
 echo      -h                : Show this message.
 echo      -f                : Publish default framework directory in ^<script_root^>\fx.
 echo      -p                : Publish utilities.
+echo      -t ^<TARGET^>     : Target framework. Default is netcoreapp2.0."
```
BruceForstall: Extra trailing quote?

sbomer: Good catch, thanks!

PR #122, opened 2017-10-31
Show when number of methods differs in base/diff

AndyAyersMS:
When multiple same-named methods exist in different numbers in base and
diff, make it clearer in the output. Also call out methods that are exclusive
to base or diff.

Allow analysis to skip reconciliation; this option is useful when
the larger diffs are all in unique methods and you want to focus in on the
methods that are common.


AndyAyersMS:
@rartemev PTAL
cc @dotnet/jit-contrib

PR #124, opened 2017-12-11
Add support for generating asm diffs using an altjit

BruceForstall:
e.g., `jit-diff --altjit protononjit.dll ...`


BruceForstall:
@dotnet/jit-contrib PTAL

PR #128, opened 2018-05-31
PMI: rework as visitor pattern to support generic type exploration

AndyAyersMS:
This change adds the ability for PMI to create instantations of some
generic types (limited to types with one or two parameters only for now).

Because of constraints it is not possible to easily tell in advance how
many instantations will result. Also in the future we will want to do
something similar for generic methods, and may want to produce different
kinds of output markup.

So this change also refactors the PMI code to use a visitor pattern for
traversing types and methods. Three concrete visitors are implemented:
one for counting types and methods, one for jitting a range of methods, and
one for jitting just a single method.

The actual strategy for instantiating generics is held in the `Worker` class.
We attempt up to 4 instantations of single parameter types and 8 of two
parameter types, with a mixture of types. There is room for enhancement here,
both in trying out suitable struct types and in looking for types that will
satisfy various constraints. A limited amount of constraint validation is
done up front.

Aside from extra methods and types seen via instantations this should be
functionally equivalent to the base version.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -602,6 +602,8 @@ class Worker
         // Only handle the very simplest cases for now
         if (genericArguments.Length > 2)
         {
+            Console.WriteLine();
+            Console.WriteLine($"Failed to instantiate {type.FullName} -- too many type parameters");
             return results;
         }
 
@@ -652,15 +654,22 @@ class Worker
                     results.Add(newType);
                 }
             }
-            catch (Exception)
+            catch (Exception e)
             {
-                // Probably missing a constraint check
+                Console.WriteLine();
+                Console.WriteLine($"TypeInstantationException {type.FullName} - {e.Message}");
             }
 
             if (instantiationCount >= instantiationLimit)
             {
                 break;
             }
+        }
+
+        if (instantiationCount == 0)
+        {
+            Console.WriteLine();
+            Console.WriteLine($"Failed to instantiate {type.FullName} -- could not find valid type substitutions");
         }
 
         return results;
@@ -673,6 +682,8 @@ class Worker
     static bool AreConstraintsSatisfied(Type type, Type parameterType)
     {
         bool areConstraintsSatisfied = true;
+
+        // Check special constraints
         GenericParameterAttributes gpa = parameterType.GenericParameterAttributes;
 
         if ((gpa & GenericParameterAttributes.ReferenceTypeConstraint) != 0)
@@ -682,7 +693,9 @@ class Worker
                 areConstraintsSatisfied = false;
             }
         }
-        else
+
+        // If all special constaints are satisfied, check type constraints
+        if (areConstraintsSatisfied)
         {
             Type[] constraints = parameterType.GetGenericParameterConstraints();
 
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -5,83 +5,236 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Numerics;
 using System.Reflection;
 
-struct MicroMethodInfo
+// This set of classes provides a way to forcibly jit a large number of methods.
+// It can be used as is or included as a component in jit measurement and testing
+// tools.
+//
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@BruceForstall hopefully not too radical of a change.
cc @dotnet/jit-contrib

src/pmi/pmi.cs
```diff
@@ -100,85 +398,323 @@ private static TimeSpan PrepareMethod(Type t, MicroMethodInfo mMI)
...
+    // It is ok if we miss some, as we catch the exception that will
+    // arise when instantating with an invalid type.
+    static bool AreConstraintsSatisfied(Type type, Type parameterType)
     {
-        int nextMethodToPrep = (methodBeingPrepped + 1);
+        bool areConstraintsSatisfied = true;
+        GenericParameterAttributes gpa = parameterType.GenericParameterAttributes;
 
-        using (var writer = new StreamWriter(File.Create("NextMethodToPrep.marker")))
+        if ((gpa & GenericParameterAttributes.ReferenceTypeConstraint) != 0)
```
CarolEidt: I'm probably missing something here, but if this condition is true, you won't check for the other constraints below. I'm not sure if `ReferenceTypeConstraint` would be "optimized away" on the type if it specifies a base type, but I would think that an interface constraint could still apply.

AndyAyersMS: I think you're right -- if this check succeeds we still should do the other checks.

It's ok for this method to be a bit sloppy, The runtime will do the real checking when PMI tries to instantiate the type and will throw an exception if the types aren't suitable. 

But it's nice if we can catch these up front to reduce the volume of unnecessary exceptions.

src/pmi/pmi.cs
```diff
@@ -100,85 +398,323 @@ private static TimeSpan PrepareMethod(Type t, MicroMethodInfo mMI)
...
-    private static int GetRMHCountOnly(Type t)
+    public int Work(string assemblyName)
     {
-        if (Object.ReferenceEquals(t, typeof(GlobalMethodHolder)))
+        Assembly assembly = LoadAssembly(Path.GetFullPath(assemblyName));
+
+        if (assembly == null)
         {
-            return GlobalMethodHolder.GlobalMethodInfoSet.Length;
+            return 102;
```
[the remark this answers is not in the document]
AndyAyersMS: This is more or less what the old code did.

We may eventually need to parameterize the Walker so we can have different failure handling mechanisms depending on how we integrate this into higher level tools.

src/pmi/pmi.cs
```diff
@@ -100,85 +398,323 @@ private static TimeSpan PrepareMethod(Type t, MicroMethodInfo mMI)
...
         }
 
-        MethodInfo[] mi = t.GetMethods(BindingFlagsForCollectingAllMethodsOrCtors);
+        List<Type> types = LoadTypes(assembly);
 
-        ConstructorInfo[] ci = t.GetConstructors(BindingFlagsForCollectingAllMethodsOrCtors);
+        if (types == null)
+        {
+            return 103;
```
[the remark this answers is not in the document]
AndyAyersMS: Same as above -- roughly what the old code did.

src/pmi/pmi.cs
CarolEidt: Thanks

src/pmi/pmi.cs
```diff
@@ -100,85 +398,323 @@ private static TimeSpan PrepareMethod(Type t, MicroMethodInfo mMI)
...
+                // If we can instantiate, prepare the methods.
+                if (newType != null)
+                {
+                    results.Add(newType);
+                }
+            }
+            catch (Exception)
+            {
+                // Probably missing a constraint check
+            }
```
BruceForstall: should we log a warning here?

AndyAyersMS: Sure.

We should also keep track of types that we fail to instantiate with our simple roster of types, since we won't jit any of those types' methods.

I'll add a note on the latter to the master todo issue.

AndyAyersMS: Updated to note types that we can't instantiate and errors when we try and instantiate.

PR #131, opened 2018-06-01
PMI: Hook up assembly resolve handler

AndyAyersMS:
Previous changes removed the bit of code to actually install the
assembly resolve handler.

Also, lay ground work for being able to re-use a visitor to visit
multiple assemblies.


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib 

Adding back a bit of code that got dropped in earlier refactoring.

PR #136, opened 2018-06-15
Fix pmi so it works properly with --altjit

AndyAyersMS:
We should always be copying/restoring via the test jit name.

### DIFF
--- a/src/jit-diff/diff.cs
+++ b/src/jit-diff/diff.cs
@@ -515,8 +515,8 @@ namespace ManagedCodeGen
 
             void InstallBaseJit()
             {
-                string realJitPath = Path.Combine(m_config.CoreRoot, m_defaultJitName);
-                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_defaultJitName);
+                string realJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
+                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_testJitName);
                 string testJitPath = Path.Combine(m_config.BasePath, m_testJitName);
                 if (m_config.Verbose)
                 {
@@ -532,8 +532,8 @@ namespace ManagedCodeGen
 
             void InstallDiffJit()
             {
-                string realJitPath = Path.Combine(m_config.CoreRoot, m_defaultJitName);
-                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_defaultJitName);
+                string realJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
+                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_testJitName);
                 string testJitPath = Path.Combine(m_config.DiffPath, m_testJitName);
                 if (m_config.Verbose)
                 {
@@ -549,8 +549,8 @@ namespace ManagedCodeGen
 
             void RestoreDefaultJit()
             {
-                string realJitPath = Path.Combine(m_config.CoreRoot, m_defaultJitName);
-                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_defaultJitName);
+                string realJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
+                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_testJitName);
                 if (m_config.Verbose)
                 {
                     Console.WriteLine($"Restoring default jit: {tempJitPath} ==> {realJitPath}");
--- a/src/jit-diff/diff.cs
+++ b/src/jit-diff/diff.cs
@@ -515,47 +515,56 @@ namespace ManagedCodeGen
 
             void InstallBaseJit()
             {
-                string realJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
-                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_testJitName);
+                string existingJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
+                string backupJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_testJitName);
                 string testJitPath = Path.Combine(m_config.BasePath, m_testJitName);
+                if (File.Exists(existingJitPath))
+                {
+                    if (m_config.Verbose)
+                    {
+                        Console.WriteLine($"Copying exsting jit: {existingJitPath} ==> {backupJitPath}");
+                    }
+                    File.Copy(existingJitPath, backupJitPath, true);
+                }
                 if (m_config.Verbose)
                 {
-                    Console.WriteLine($"Copying default jit: {realJitPath} ==> {tempJitPath}");
+                    Console.WriteLine($"Copying in the test jit: {testJitPath} ==> {existingJitPath}");
                 }
-                File.Copy(realJitPath, tempJitPath, true);
-                if (m_config.Verbose)
-                {
-                    Console.WriteLine($"Copying in the test jit: {testJitPath} ==> {realJitPath}");
-                }
-                File.Copy(testJitPath, realJitPath, true);
+                File.Copy(testJitPath, existingJitPath, true);
             }
 
             void InstallDiffJit()
             {
-                string realJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@CarolEidt PTAL (and if possible, give this a try and make sure it works for you).
cc @dotnet/jit-contrib

CarolEidt:
This gets a bit further - it compiles two methods before getting the `Assert failure(PID 31772 [0x00007c1c], Thread: 32764 [0x7ffc]): !"Recursion in CLRException::GetThrowable"`. It seems to be listing each method twice - and it doesn't appear that we log the target OS in the assembly listing, so I can't tell off hand whether it's doing a disassembly listing both for the main & altjits.

src/jit-diff/diff.cs
```diff
@@ -515,8 +515,8 @@ protected override List<string> ConstructArgs(List<string> commandArgs, string c
 
             void InstallBaseJit()
             {
-                string realJitPath = Path.Combine(m_config.CoreRoot, m_defaultJitName);
-                string tempJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_defaultJitName);
+                string realJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
```
BruceForstall: What if the install doesn't have a jit by this name? Seems like this could happen for some, maybe new, altjit.

Seems like "tempJitPath" should be "backupJitPath".

AndyAyersMS: Yeah, probably warrants some more complex logic.

AndyAyersMS:
Unfortunately I can't repro the failure you see.

After you pulled down the changes  for this PR did you rebuilt jitutils? (via `build -p` or `bootstrap.cmd`)

It's also possible the older jitutils build messed up your test overlay dirs, so you might want to rerun `runtest.cmd ... generatelayoutonly` in the base and diff repos to reset the overlays to their original state.

With `linuxnonjit` I hit an assert in corelib:
```
D:\repos\coreclr>jit-diff diff --base --base_root d:\repos\coreclr2 --pmi --diff --altjit linuxnonjit.dll --build Checked
...
Assert failure(PID 23120 [0x00005a50], Thread: 12036 [0x2f04]): Assertion failed 'tree->gtHasReg()' in 'System.Runtime.InteropServices.WindowsRuntime.ConstantSplittableMap`2[Vector`1,Vector`1][System.Numerics.Vector`1[System.Single],System.Numerics.Vector`1[System.Single]]:TryGetValue(struct,byref):bool:this' (IL size 85)
    File: d:\repos\coreclr\src\jit\codegenlinear.cpp Line: 1204
    Image: D:\repos\coreclr\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\corerun.exe
```

AndyAyersMS:
And I agree it would be nice for the listings to be a bit more expressive -- also we should push that information out into the diff summary since we tend to keep just that text and it's easy to forget what it was comparing.

CarolEidt:
@AndyAyersMS - thanks! I can now repro that failure. I'll have a look. It's possibly an altjit-specific issue, since it's an instantiation over `Vector<Single>`

AndyAyersMS:
I've run your change over all the Fx assemblies now and don't see any diffs. Trying all the tests now. It is slow going as each method gets jitted 2x and (for now) I'm also using a checked build as the host.

I'm using COMPlus_AltJitSkiopOnAssert=1 and COMPlus_ContinueOnAssert=1 to try and plow ahead. See maybe 50 instances of the assert above across Fx assemblies.

AndyAyersMS:
All tests diffs did not call out any diffs either -- even on your new test case.

It turns out there are diffs, but jit-analyze doesn't flag methods with diffs if they don't actually change the length of the method. In this case just some immediate offsets changed:

```asm
;;; before
       mov      word  ptr [rsp], ax
       mov      dword ptr [rsp+02H], edi
       vmovss   dword ptr [rsp+06H], xmm0
       mov      qword ptr [rsp+0AH], rsi
       call     Test:foo(struct):int

;;; after
       mov      word  ptr [rsp], ax
       mov      dword ptr [rsp+04H], edi
       vmovss   dword ptr [rsp+08H], xmm0
       mov      qword ptr [rsp+10H], rsi
       call     Test:foo(struct):int
```
So the upshot is we can't easily tell for a change like this if it caused diffs or not.

I will have to look at jit-analyze more closely to see if there is some way to flag methods with assembly diffs but not size diffs.

CarolEidt:
> I will have to look at jit-analyze more closely to see if there is some way to flag methods with assembly diffs but not size diffs.

Right - currently I believe that you get a message like "found files with textual diffs", but it would be really good to distinguish between constants that differ (which I believe we still have even with `COMPlus_JitDiffableDasm`) or different registers allocated, versus something like this (though distinguishing offsets from runtime-dependent constants might be tricky)

CarolEidt:
> See maybe 50 instances of the assert above across Fx assemblies.

I believe I have a fix for the case in SPC.dll. I'm not sure why we haven't hit this, but I'm assuming it's because of the way pmi instantiates types that we otherwise wouldn't see. If you have a small test case, that would be great, but I'll work on extracting one.

AndyAyersMS:
@BruceForstall last commit should hopefully address your comment.
cc @dotnet/jit-contrib

AndyAyersMS:
I'll probably update jit-analyze to track files with text diffs that don't have size diffs, and at least for now print out the top N of those. For instance via `--numstat` or similar.

Later perhaps we can capture diff ranges for these no-size-diff files and intersect those with the function ranges and perhaps provide more detailed info on which functions were impacted.

For your bugfix change running over Pri0 tests here's what we would have seen, more or less:
```
D:\repos\coreclr>git diff --stat --exit-code  --no-index bin\diffs\dasmset_6\base bin\diffs\dasmset_6\diff
 .../coverage/oldtests/cse2_cs_do/cse2_cs_do.dasm"  |  4 +-
 .../coverage/oldtests/cse2_cs_ro/cse2_cs_ro.dasm"  |  4 +-
 .../25params/25param2c_il_d/25param2c_il_d.dasm"   |  4 +-
 .../V1-M12-Beta2/b71120/b71120/b71120.dasm"        |  4 +-
 .../V1.2-Beta1/b219940/b219940/b219940.dasm"       |  4 +-
 .../GitHub_17585/GitHub_17585/GitHub_17585.dasm"   |  4 +-
 .../GitHub_18482/GitHub_18482/GitHub_18482.dasm"   |  6 +-
 .../JIT/jit64/mcc/interop/mcc_i06/mcc_i06.dasm"    |  4 +-
 .../JIT/jit64/mcc/interop/mcc_i16/mcc_i16.dasm"    |  4 +-
 .../JIT/jit64/mcc/interop/mcc_i36/mcc_i36.dasm"    |  4 +-
 .../JIT/jit64/mcc/interop/mcc_i56/mcc_i56.dasm"    |  4 +-
 .../JIT/jit64/mcc/interop/mcc_i66/mcc_i66.dasm"    |  4 +-
 .../JIT/jit64/mcc/interop/mcc_i76/mcc_i76.dasm"    |  4 +-
 .../JIT/jit64/mcc/interop/mcc_i86/mcc_i86.dasm"    |  4 +-
 .../TailcallVerifyWithPrefix.dasm"                 | 76 +++++++++++-----------
 15 files changed, 67 insertions(+), 67 deletions(-)
```
Of these most are asserts or complaints about varargs not being implemented on unix (I suppose PMI might also need to start paying attention to exclusion lists, but that seems tricky).

These two tests had actual codegen diffs:
```
 .../coverage/oldtests/cse2_cs_do/cse2_cs_do.dasm"  |  4 +-
 .../coverage/oldtests/cse2_cs_ro/cse2_cs_ro.dasm"  |  4 +-
```

src/jit-diff/diff.cs
```diff
@@ -515,47 +515,56 @@ protected override List<string> ConstructArgs(List<string> commandArgs, string c
...
+                string existingJitPath = Path.Combine(m_config.CoreRoot, m_testJitName);
+                string backupJitPath = Path.Combine(m_config.CoreRoot, "backup-" + m_testJitName);
                 string testJitPath = Path.Combine(m_config.BasePath, m_testJitName);
-                if (m_config.Verbose)
+                if (File.Exists(existingJitPath))
                 {
-                    Console.WriteLine($"Copying default jit: {realJitPath} ==> {tempJitPath}");
+                    if (m_config.Verbose)
+                    {
+                        Console.WriteLine($"Copying exsting jit: {existingJitPath} ==> {backupJitPath}");
```
BruceForstall: typo: exsting

PR #137, opened 2018-06-17
jit-analyze: show top N files with text diffs but no size diffs

AndyAyersMS:
Start working towards highlighting files where codegen (potentially) differs
but the overall method size stays the same.

In this initial version jit-analyze will list the top N files files that have
zero size diffs but have some text diff, ordered by decreasing number of diff
lines.

N here is controlled by the --count parameter like the other "top N" sections
of the diff summary and defaults to 5.

Users can then manually inspect diffs in those files to see what, if any, code
has changed in the methods in the file.

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -148,7 +148,8 @@ namespace ManagedCodeGen
 
         public class FileDelta
         {
-            public string path;
+            public string basePath;
+            public string diffPath;
             public int baseBytes;
             public int diffBytes;
             public int deltaBytes;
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@CarolEidt PTAL
cc @dotnet/jit-contrib 

For the change in dotnet/coreclr#18483, `jit-analyze` now prints:
```
Summary:
(Note: Lower is better)

Total bytes of diff: 0 (0.00% of base)

0 total files with size differences (0 improved, 0 regressed), 2463 unchanged.

0 total methods with size differences (0 improved, 0 regressed), 166637 unchanged.

4 files had text diffs but not size diffs.
JIT\Regression\JitBlue\GitHub_18482\GitHub_18482\GitHub_18482.dasm had 12 diffs
JIT\Directed\coverage\oldtests\cse2_cs_do\cse2_cs_do.dasm had 8 diffs
JIT\Directed\coverage\oldtests\cse2_cs_ro\cse2_cs_ro.dasm had 8 diffs
JIT\Regression\CLR-x86-JIT\V1-M12-Beta2\b71120\b71120\b71120.dasm had 0 diffs
```
The first 3 are the known diffs from the change (some immediate offset values differ). The last file gets parsed as a binary file by git (not sure why just yet). It has diffs because it triggers two NYI assertions related to varargs and these interpolate differently into the disassembly listings.

PR #138, opened 2018-06-17
pmi: add -TIME option and make -QUIET mode more quiet

AndyAyersMS:
The -QUIET suffix now produces just one output line per assembly with no timing
data. It is intended for use by upstack tools like `jit-diffs` to reduce the
volume of diff text and reduce spurious diffs caused by PMI messages.

Add a new -TIME suffix to print per-assembly times, even in -quiet mode. This
is useful when using PMI as a jit throughput tool.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -52,9 +52,8 @@ class Visitor
     {
     }
 
-    public virtual void UninstantiableType(Type t)
+    public virtual void UninstantiableType(Type type, string reason)
     {
-
     }
 
     public virtual void StartType(Type type)
@@ -118,9 +117,9 @@ class CounterBase : Visitor
         typeCount++;
     }
 
-    public override void UninstantiableType(Type t)
+    public override void UninstantiableType(Type type, string reason)
     {
-        base.UninstantiableType(t);
+        base.UninstantiableType(type, reason);
         uninstantiableTypeCount++;
     }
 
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib 

Adapt PMI to be more useful for exploring jit throughput.

src/pmi/pmi.cs
```diff
@@ -1031,13 +1038,17 @@ public static int Main(string[] args)
                     }
                 }
 
-                if ((command == "PREPALL") || (command == "PREPALL-QUIET"))
+                bool all = command.IndexOf("ALL") > 0;
+                bool verbose = !(command.IndexOf("QUIET") > 0);
+                bool time = verbose || command.IndexOf("TIME") > 0;
+
```
BruceForstall: Seems like we should more to a more "standard" argument passing model. It looks like this parser will support something like `PREPALL-QUIETFOOBARTIME`, which is weird.

src/pmi/pmi.cs
```diff
@@ -199,9 +200,10 @@ public override void FinishAssembly(Assembly assembly)
             $"Completed assembly {assemblyName} - #types: {typeCount}, #methods: {methodsPrepared}, " +
             $"skipped types: {uninstantiableTypeCount}, skipped methods: {uninstantiableMethodCount}");
 
-        if (_verbose)
+        if (_time || _verbose)
         {
-            Console.WriteLine($", elapsed ms: {elapsed.TotalMilliseconds:F2}");
+            string prefix = _time && !_verbose ? "" : ", ";
+            Console.WriteLine($"{prefix}Time: {elapsed.TotalMilliseconds:F2}ms");
```
BruceForstall: So in the `_time && !_verbose` case there is no ", "? doesn't that just jam it against the skipped methods count?

src/pmi/pmi.cs
AndyAyersMS: Agree we probably should rework this, but it's not high on my list of todos.

src/pmi/pmi.cs
AndyAyersMS: Yeah, not sure what I was thinking here. Will fix.

Issue #140, opened 2018-06-20
PMI Usage message is incorrect


BruceForstall:
The dotnet CLI-version of the usage message prints "dotnet.exe" as the tool, which is confusing:

```
c:\gh\jitutils>where pmi
c:\gh\jitutils\bin\pmi.bat

c:\gh\jitutils>pmi -?
Usage:

  dotnet.exe Count PATH_TO_ASSEMBLY
      Count the number of types and methods in an assembly.

  dotnet.exe PrepOne PATH_TO_ASSEMBLY INDEX_OF_TARGET_METHOD
      JIT a single method, specified by a method number.

  dotnet.exe PrepAll PATH_TO_ASSEMBLY [INDEX_OF_FIRST_METHOD_TO_PROCESS]
      JIT all the methods in an assembly. If INDEX_OF_FIRST_METHOD_TO_PROCESS is specified, it is the first
      method compiled, followed by all subsequent methods.

  dotnet.exe DriveAll PATH_TO_ASSEMBLY
      The same as PrepAll, but is more robust. While PrepAll will stop at the first JIT assert, DriveAll will
      continue by skipping that method.

Environment variable PMIPATH is a semicolon-separated list of paths used to find dependent assemblies.

For Prepall and Prepone, optional suffixes will change output behavior:
   -Quiet will suppress in-progress messages for type and method exploration
   -Time will always show elapsed times, even in -Quiet mode
```

This is from:
```
private static int Usage()
{
    string exeName = System.Diagnostics.Process.GetCurrentProcess().MainModule.FileName; // get the current full path name of PMI.exe
    exeName = System.IO.Path.GetFileName(exeName); // strip off the path; just use the EXE name.
    Console.WriteLine(
        "Usage:\r\n"
        + "\r\n"
        + "  " + exeName + " Count PATH_TO_ASSEMBLY\r\n"
```

BruceForstall:
@AndyAyersMS

AndyAyersMS:
The exe name is indeed dotnet.exe. There is no implicit CLR bootstrapper for core apps like there is for full framework apps.

IIRC there is some way to publish so you end up with a suitably named bootstrapping exe (perhaps by publishing for specific runtimes?). So maybe we should switch to that.

BruceForstall:
Even if the name is dotnet.exe, the command line needs to include the pmi.dll.

kunalspathak:
Looks like this is fixed in #144.

PR #144, opened 2018-06-20
Some minor usability tweaks in jitutils

AndyAyersMS:
Have `jit-diff` pass a description string down to `jit-analyze` so the
summary can describe what was being diffed. Capture the arch that was
actually used and add it to the description.

Have `jit-analyze` accept that description string and display it in the
summary.

Remove a diagnostic "adding" message inadvertenly left in `jit-analyze`'s
processing of files with text diffs.

Don't show final file diff (but no size diff) message in the analysis
summary if there were no such files (which is commonly the case).

Update Usage for `pmi` to indicate you must pass `pmi.dll` as the first arg
to your launcher if you launched via `dotnet` or `corerun`. This should
partially address #140.

Add a bit more EH to `jit-dasm-pmi` in a attempt to reduce the volume of
dotnet failure popup dialogs (seen when altjit asserts cause crashes). More
is needed here to make this fully robust.

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -29,6 +29,7 @@ namespace ManagedCodeGen
             private string _json;
             private string _tsv;
             private bool _noreconcile = false;
+            private string _note;
 
             public Config(string[] args)
             {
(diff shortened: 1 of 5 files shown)


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib

AndyAyersMS:
Yeah, we should make the repro paths easier. 

I would like all the failure details to be logged to a file (and only briefly summarized by default in standard output). Once we do that we can make them look more or less like exactly what you need type in to repro the issue.

PR #150, opened 2018-06-24
Progress bar and SIMD size

AndyAyersMS:
Set the SIMD size if the jit looks like it is cross-targeting. Currently
just checking altjits for "non".

Add simple progress bar to jit-diffs that shows up if you are running with
output to the console.

### DIFF
--- a/src/jit-dasm-pmi/jit-dasm-pmi.cs
+++ b/src/jit-dasm-pmi/jit-dasm-pmi.cs
@@ -139,7 +139,7 @@ namespace ManagedCodeGen
         public bool UseFileName { get { return (_fileName != null); } }
         public bool DumpGCInfo { get { return _dumpGCInfo; } }
         public bool DoVerboseOutput { get { return _verbose; } }
-        public bool CopyJit {  get { return !_noCopyJit; } }
+        public bool CopyJit { get { return !_noCopyJit; } }
         public string CorerunExecutable { get { return _corerunExe; } }
         public string JitPath { get { return _jitPath; } }
         public string AltJit { get { return _altjit; } }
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib 

jitdiffprogressbar

src/jit-dasm-pmi/jit-dasm-pmi.cs
```diff
@@ -483,6 +483,14 @@ void GenerateAsmInternal()
                         generateCmd.EnvironmentVariable("COMPlus_AltJit", "*");
                         generateCmd.EnvironmentVariable("COMPlus_AltJitName", _altjit);
 
+                        // If this looks like a cross-targeting altjit, fix the SIMD size.
+                        // Here's one place where rationalized jit naming would be nice.
+                        if (_altjit.IndexOf("nonjit") > 0)
```
BruceForstall: I guess this works... seems fragile...

src/jit-diff/diff.cs
```diff
@@ -27,6 +20,58 @@ public enum DasmWorkKind
         Diff
     }
 
+    public class ProgressBar
```
BruceForstall: Fancy :-)

IMO, this might be "too much" progress output. seems like maybe once per second would be fine.

AndyAyersMS: I tried it at 5s and 1s initially and liked 0.1s better... Especially early, on when it seems not much is happening as we do some of the larger assemblies first.

briansull: Maybe it should be progressive, 0.1 for the first 2 seconds, then 1s seconds after that?

AndyAyersMS: Give this a try, and if it's annoying, we can work on improving it.

AndyAyersMS:
Grr, my changes are not playing nice with one another...
```
- Finished 0/1 Base 0/1 Diff [0.1 sec]Setting SIMD Length to 16
/ Finished 1/1 Base 0/1 Diff [0.5 sec]
- Finished 1/1 Base 0/1 Diff [0.6 sec]Setting SIMD Length to 16
- Finished 1/1 Base 1/1 Diff [1.0 sec]
```
I suppose the "setting SIMD bit" could be something we put under verbose or in some RSP file.

PR #152, opened 2018-06-26
Fix a couple of quirks with status bar

AndyAyersMS:
When using altjit, don't print the SIMD message by default.
Base runs are the last set when there aren't any diff runs.


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib 

Second diff is a bit hard to see, it's adding a `!`.

PR #154, opened 2018-06-30
Add More Benchmarks and Tsv option

CarolEidt:
The path to BenchI and BenchF has changed, so only BenchmarksGame was being used with the --benchmarks option.
Also, I like to have the tsv file generated, and would prefer to save having the extra analyze step, so add that as a pass-through option.


CarolEidt:
@dotnet/jit-contrib PTAL

PR #155, opened 2018-07-11
jit-pmi-diff: disable tiering by default

AndyAyersMS:
Turn off tiering by default for when invoking PMI. Can enable via -tiering.

Also, fix a couple of spelling errors.

### DIFF
--- a/src/jit-dasm-pmi/jit-dasm-pmi.cs
+++ b/src/jit-dasm-pmi/jit-dasm-pmi.cs
@@ -42,6 +42,7 @@ namespace ManagedCodeGen
         private bool _dumpGCInfo = false;
         private bool _noCopyJit = false;
         private bool _verbose = false;
+        private bool _tiering = false;
 
         public Config(string[] args)
         {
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib

src/jit-dasm-pmi/jit-dasm-pmi.cs
```diff
@@ -147,6 +149,7 @@ private void Validate()
         public IReadOnlyList<string> PlatformPaths { get { return _platformPaths; } }
         public string FileName { get { return _fileName; } }
         public IReadOnlyList<string> AssemblyList { get { return _assemblyList; } }
+        public bool Tiering => _tiering;
```
BruceForstall: Doesn't this syntax make the property read/write, when all we really want is to make it readable?

AndyAyersMS: No, this is just a getter, to be able to set you'd need something like:
```
   public Bool Tiering
   {
      get => _tiering;
      set => _tiering = value;
   }
```

AndyAyersMS:
Tiering doesn't impact crossgen codegen, so jit-dasm doesn't care one way or the other... or are you suggesting something else?

BruceForstall:
Oh, forgot about that. ok.

Issue #157, opened 2018-07-19
jit-diff --pmi fewer number of jitted methods than jit-diff --crossgen for S.P.C.dll


echesakovMSFT:
**PMI**
```
D:\git>E:\git\jitutils\bin\jit-diff.bat diff --base E:\git\coreclr-base\bin\Product\Windows_NT.x64.Checked --output D:\JitDiff --pmi --core_root E:\git\coreclr-base\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root
No assemblies specified; defaulting to corelib
Beginning PMI Diffs for System.Private.CoreLib.dll
- Finished 1/1 Base 0/0 Diff [17.7 sec]
Completed PMI Diffs for System.Private.CoreLib.dll in 17.73s
Diffs (if any) can be viewed by comparing: D:\JitDiff\dasmset_9\base D:\JitDiff\dasmset_9\diff

echesako@echesako1:/mnt/d/JitDiff$ cat dasmset_9/base/System.Private.CoreLib.dasm | grep "Assembly listing for method" | wc -l
4967
```

**Crossgen**
```
D:\git>E:\git\jitutils\bin\jit-diff.bat diff --base E:\git\coreclr-base\bin\Product\Windows_NT.x64.Checked --output D:\JitDiff --crossgen E:\git\coreclr-base\bin\Product\Windows_NT.x64.Checked\crossgen.exe --core_root E:\git\coreclr-base\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root
No assemblies specified; defaulting to corelib
Beginning Crossgen Diffs for System.Private.CoreLib.dll
| Finished 1/1 Base 0/0 Diff [60.8 sec]
Completed Crossgen Diffs for System.Private.CoreLib.dll in 60.83s
Diffs (if any) can be viewed by comparing: D:\JitDiff\dasmset_10\base D:\JitDiff\dasmset_10\diff

echesako@echesako1:/mnt/d/JitDiff$ cat dasmset_10/base/System.Private.CoreLib.dasm | grep "Assembly listing for method" | wc -l
28267
```

echesakovMSFT:
Closed via #158

PR #158, opened 2018-07-19
Add COMPlus_ZapDisable=1 to output disasm for pre-jitted methods

echesakovMSFT:
To address #157


echesakovMSFT:
@AndyAyersMS PTAL
cc @dotnet/jit-contrib

src/jit-dasm-pmi/jit-dasm-pmi.cs
```diff
@@ -482,6 +482,7 @@ void AddEnvironmentVariable(string varName, string varValue)
                     AddEnvironmentVariable("COMPlus_JitUnwindDump", "*");
                     AddEnvironmentVariable("COMPlus_JitEHDump", "*");
                     AddEnvironmentVariable("COMPlus_JitDiffableDasm", "1");
+                    AddEnvironmentVariable("COMPlus_ZapDisable", "1");
```
AndyAyersMS: For future-proofing we should also set `COMPlus_ReadyToRun=0` to disable R2R method loads too.

CarolEidt:
@echesakovMSFT - thanks for noticing this and tracking it down!

PR #159, opened 2018-07-25
PMI: Improve constraint checking logic

AndyAyersMS:
The logic to check for generic parameter type constraints was wrong.
If the constraint it itself generic then just skip checking and let
the runtime figure it out.

Handle the other classes of special constraint. Fix the check for the
class/interface constraint.

Add `long` to the types we try and use for instantiations since it
stresses x86.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -784,7 +784,7 @@ class Worker
         }
 
         // Types we will use for instantiation attempts.
-        Type[] typesToTry = new Type[] { typeof(object), typeof(int), typeof(double), typeof(Vector<float>) };
+        Type[] typesToTry = new Type[] { typeof(object), typeof(int), typeof(double), typeof(Vector<float>), typeof(long) };
 
         // To keep things sane, we won't try and instantiate too many copies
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
cc @dotnet/jit-contrib 

Increases "coverage" from PMI somewhat, eg:

```
;; before
Completed assembly System.Private.Xml - #types: 1686, #methods: 18044, skipped types: 22, skipped methods: 221
Completed assembly VectorExp_ro - #types: 2, #methods: 15, skipped types: 1, skipped methods: 9

;; after
Completed assembly System.Private.Xml - #types: 1707, #methods: 18246, skipped types: 22, skipped methods: 221
Completed assembly VectorExp_ro - #types: 5, #methods: 21, skipped types: 1, skipped methods: 9
```

src/pmi/pmi.cs
```diff
@@ -875,23 +890,17 @@ static bool AreConstraintsSatisfied(Type type, Type parameterType)
 
             foreach (Type c in constraints)
             {
-                if (c.IsClass)
+                // If the constrant is also generic, just bail on checks.
```
BruceForstall: typo: constrant

PR #160, opened 2018-07-25
PMI: instantiate generic methods

AndyAyersMS:
Also fix issue instantiating two-parameter generic types where we
would only ever try the last suggested type for the second type param.

Also fix accounting for instantiated types and methods.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -754,11 +754,28 @@ class Worker
         bool keepGoing = true;
         foreach (MethodBase methodBase in GetMethods(type))
         {
-            visitor.StartMethod(type, methodBase);
-            keepGoing = visitor.FinishMethod(type, methodBase);
-            if (!keepGoing)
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
cc @dotnet/jit-contrib 

Gets us "coverage" for generic methods and fixes some warts.

src/pmi/pmi.cs
```diff
@@ -820,12 +824,21 @@ private List<Type> GetInstances(Type type)
...
                 {
-                    newType = type.MakeGenericType(firstType, secondType);
+                    foreach (Type secondType in typesToTry)
+                    {
+                        if (!AreConstraintsSatisfied(secondType, genericArguments[1]))
+                        {
+                            continue;
+                        }
+
+                        newType = type.MakeGenericType(firstType, secondType);
```
BruceForstall: should there be a try/catch around this, so we try the rest of the "typesToTry" for the second argument?

PR #162, opened 2018-08-20
Avoid re-reading .dasm files

mikedn:
Isn't LINQ nice when re-evaluating a query results in re-reading gigabytes of data from disk?


mikedn:
@BruceForstall @AndyAyersMS This speeds-up jit-analyze for a x64 fx diff by something like 5x (it was 130s, now it's 24s).

In the ARM case it was even worse on my machine, the .dasm files are larger and apparently do not fit in the filesystem cache. It was already taking a long time to display "Total bytes of diff: -280 (0.00% of base)" and then it was taking forever to display the rest.

It may be possible to get below 24s by reading the files in a smarter way (e.g. map the file), maybe I'll look at that in the future. Though around half of that 24s is running `git diff` so perhaps there's not a lot more to gain by improving jit-analyze.

AndyAyersMS:
Nice....

PR #169, opened 2018-10-11
Use COMPlus_JitStdOutFile to speed-up disassembly

### LINKED ISSUE #161: Improve jit-dasm performance
It looks like jit-dasm consumes a lot of CPU:
jit-dasm

And that's when running it sequentially. In parallel you end up with multiple instances of jit-dasm that each take 17% CPU...

As far as I can tell this is due to routing the crossgen output through it. It would be better if the crossgen output goes straight to a file. Unfortunately there doesn't seem to be an easy way to do that:
* the current approach is dictated by `Microsoft.DotNet.Cli.Utils.Command` works, it doesn't appear to offer any other option
* `System.Diagnostics.Process` doesn't seem to be any better, it handles redirection internally as well and gives you the pipe streams, it doesn't give you the opportunity to provide your own stream.
* PInvoke-ing native OS functionality would do the job but it's kind of hairy, especially if you want to make this work on both Windows and Linux.
* Another option would be to modify crossgen to accept a command line option (or env var?) telling it where to write the disassembly. After all, native compilers do offer such options (e.g. MSVC's `/Fa`). But what about PMI? I'd guess the env var would work for it too but not the command line option.
* Yet another option would be to take advantage of the native shell redirection capabilities. Starting `crossgen foobar.dll > out.dasm` should probably do it. It's still platform dependent but the only difference is probably `cmd -c blah` vs. `sh -c blah`.

mikedn:
Relying on stdout redirection is not very efficient. The output is piped through jit-dasm rather than being written directly to the file and jit-dasm ends up consuming a lot of CPU. Also, the JIT disables stdio buffering when writing to the stdout and that makes things even slower.

Fixes #161

### DIFF
--- a/src/jit-dasm-pmi/jit-dasm-pmi.cs
+++ b/src/jit-dasm-pmi/jit-dasm-pmi.cs
@@ -530,21 +530,23 @@ namespace ManagedCodeGen
                     {
                         // Generate path to the output file
                         var assemblyFileName = Path.ChangeExtension(assembly.Name, ".dasm");
-                        var path = Path.Combine(_rootPath, assembly.OutputPath, assemblyFileName);
+                        var dasmPath = Path.Combine(_rootPath, assembly.OutputPath, assemblyFileName);
+                        var logPath = Path.ChangeExtension(dasmPath, ".log");
 
-                        PathUtility.EnsureParentDirectoryExists(path);
+                        PathUtility.EnsureParentDirectoryExists(dasmPath);
 
-                        // Redirect stdout/stderr to disasm file and run command.
-                        using (var outputStream = System.IO.File.Create(path))
+                        generateCmd.EnvironmentVariable("COMPlus_JitStdOutFile", dasmPath);
+
+                        // Redirect stdout/stderr to log file and run command.
+                        using (var outputStreamWriter = File.CreateText(logPath))
                         {
-                            using (var outputStreamWriter = new StreamWriter(outputStream))
-                            {
-                                // Forward output and error to file.
-                                generateCmd.ForwardStdOut(outputStreamWriter);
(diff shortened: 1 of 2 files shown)


mikedn:
cc @AndyAyersMS @BruceForstall 

This adds support for `COMPlus_JitStdOutFile` and redirects the stdout/err to a .log file. The only remaining question is what to do about older version of the JIT that do not support `COMPlus_JitStdOutFile`.

mikedn:
Interestingly, the log file for corelib pmi contains:
```
…
InvalidProgramException System.Runtime.Intrinsics.X86.Sse42::CompareExplicitLengthUnitMask
InvalidProgramException System.Runtime.Intrinsics.X86.Sse42::CompareExplicitLengthUnitMask
Completed assembly System.Private.CoreLib - #types: 2301, #methods: 27387, skipped types: 24, skipped methods: 425
```
Perhaps it's good having these in a separate file, I wouldn't have noticed these in the large .dasm file. Same for crossgen, there are a bunch of such exceptions when compiling async methods across FX (AFAIR there's an issue opened for that).

AndyAyersMS:
For PMI you could notice if the dasm file is empty and issue a warning. And then perhaps rename the log file to the dasm file, since that's where the dasm will have ended up if redirection didn't work.

mikedn:
> For PMI you could notice if the dasm file is empty and issue a warning. And then perhaps rename the log file to the dasm file, since that's where the dasm will have ended up if redirection didn't work.

Actually it works for crossgen as well, if the JIT does not support `COMPlus_JitStdOutFile` then the file won't be created at all. Thanks!

mikedn:
Updated to deal with missing JIT support for `COMPlus_JitStdOutFile`.

Issue #171, opened 2018-10-12
PMI asm diffs corrupted by timer


BruceForstall:
ASM diffs are getting corrupt output when running `jit-diff diff --corelib –pmi` resulting in incorrect diffs report. Managed code is executed (and causes jitting) on another thread and the dasm output from that thread gets interleaved with pmi dasm.

Eugene tracked his down to tracing infrastructure polling that happens even if tracing is disabled: 

https://github.com/dotnet/coreclr/blob/457148064c836daa746542f1559b7958d6420c22/src/System.Private.CoreLib/src/System/Diagnostics/Eventing/EventPipeController.cs#L146

This line recently changed the polling interval when tracing is disabled from 10 seconds to 5 seconds; that’s probably why we started seeing the problem:

https://github.com/dotnet/coreclr/blob/457148064c836daa746542f1559b7958d6420c22/src/System.Private.CoreLib/src/System/Diagnostics/Eventing/EventPipeController.cs#L40

It appears there is no way to disable the tracing polling.

One workaround is adding Thread.Sleep(6000) at the beginning of pmi Main method (so that the first polling happens before we start pmi-ing) but we need a better solution.

BruceForstall:
Fixed by https://github.com/dotnet/jitutils/issues/172

PR #172, opened 2018-10-13
Work around a jit-diff pmi output corruption.

erozenfeld:
I tracked down the corruption to tracing infrastructure polling that happens even if tracing is disabled:

https://github.com/dotnet/coreclr/blob/457148064c836daa746542f1559b7958d6420c22/src/System.Private.CoreLib/src/System/Diagnostics/Eventing/EventPipeController.cs#L146

This line recently changed the polling interval when tracing is disabled from 10 seconds to 5 seconds; that’s probably why we started seeing the problem:

https://github.com/dotnet/coreclr/blob/457148064c836daa746542f1559b7958d6420c22/src/System.Private.CoreLib/src/System/Diagnostics/Eventing/EventPipeController.cs#L40

The workaround is to force jitting of TimerCallback infrastructure before calling PMI on methods.

This unblocks pmi diff testing until we find a better general solution for these kinds of problems.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Linq;
 using System.Numerics;
 using System.Reflection;
+using System.Threading;
 
 // This set of classes provides a way to forcibly jit a large number of methods.
 // It can be used as is or included as a component in jit measurement and testing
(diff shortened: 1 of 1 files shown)


erozenfeld:
@AndyAyersMS @BruceForstall @dotnet/jit-contrib PTAL

BruceForstall:
Closes https://github.com/dotnet/jitutils/issues/171

PR #178, opened 2018-10-19
Allow --test-root to point to an arbitrary tree

BruceForstall:
If you point at an arbitrary tree, and pass --tests, then generate
asm for all assemblies in the tree. This allows you to generate
asm for just a subset of the test tree by passing that as the root.

Whether we are looking at the test root or an arbitrary subtree is
determined by looking for the test tree root subdirectories we
expect to find (JIT and Interop).

### DIFF
--- a/src/jit-diff/diff.cs
+++ b/src/jit-diff/diff.cs
@@ -426,22 +426,53 @@ namespace ManagedCodeGen
                 if (config.Benchmarks || config.DoTestTree)
                 {
                     string basepath = config.Benchmarks ? Utility.CombinePath(config.TestRoot, s_benchmarksPath) : config.TestRoot;
-                    foreach (var dir in config.Benchmarks ? s_benchmarkDirectories : s_testDirectories)
+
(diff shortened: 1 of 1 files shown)


BruceForstall:
I was using this to pass, e.g., `--test_root F:\gh\coreclr10\bin\tests\Windows_NT.x64.Release\JIT\jit64\localloc`.

Seems a bit of a hack to check for the existing test tree roots. Maybe I should add "--tree_root" and "--tree" options (similar to `--test_root` and `--tests` that we currently have, but it would always run everything in the subtree. And there would be no default for `--test_root`.

Comments?

@AndyAyersMS @dotnet/jit-contrib

AndyAyersMS:
`--assembly` is kind of like this too.

Seems like it would be nice to have a general purpose option that can point at a file or dir (and if the latter, recursively enumerates the assemblies), as well as built-in shortcuts for useful test tree subsets like the benchmarks and the "usual" set of tests.

And perhaps can be specified more than once to allow combinations.

PR #179, opened 2018-10-22
Extend jit-diff `--assembly` argument

BruceForstall:
Two changes:
1. Allow you to specify multiple `--assembly` arguments; diffs will happen
for each specified assembly.
2. Allow an argument to `--assembly` to be a directory. In this case,
the directory tree will be searched for assemblies to diff.

E.g.,
```
jit-diff diff --pmi --base --diff --base_root f:\gh\coreclr12 --arch x64 --build Checked --assembly F:\gh\coreclr10\bin\tests\Windows_NT.x64.Release\JIT\jit64\localloc --assembly F:\gh\coreclr10\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Collections.dll --assembly F:\gh\coreclr10\bin\tests\Windows_NT.x64.Checked\Tests\Core_Root\System.Collections.Specialized.dll
```

### DIFF
--- a/src/jit-diff/diff.cs
+++ b/src/jit-diff/diff.cs
@@ -279,10 +279,11 @@ namespace ManagedCodeGen
                     needPrefix = true;
                 }
 
-                if (config.AssemblyName != null)
+                foreach (string assembly in config.AssemblyList)
                 {
                     if (needPrefix) diffString += ", ";
-                    diffString += Path.GetFileName(config.AssemblyName);
+                    diffString += assembly;
+                    needPrefix = true;
                 }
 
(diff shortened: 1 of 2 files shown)


BruceForstall:
@AndyAyersMS What do you think about this, instead of https://github.com/dotnet/jitutils/pull/178?

BruceForstall:
@dotnet/jit-contrib fyi--

Two changes to jit-diff:

1. Allow you to specify multiple --assembly arguments; diffs will happen
for each specified assembly.
2. Allow an argument to --assembly to be a directory. In this case,
the directory tree will be searched for assemblies to diff.

PR #180, opened 2018-10-31
Set environment variables to assert on NO_WAY and noway_assert

BruceForstall:
We don't want to allow JIT fallback to MinOpts when generating
asm diffs.


BruceForstall:
cc @dotnet/jit-contrib

BruceForstall:
cc @dotnet/jit-contrib

PR #181, opened 2018-11-01
WIP Allow building for netcoreapp3.0

BruceForstall:
You can also choose netcoreapp2.1. Unfortunately, using a 2.1
CLI SDK fails to build, even when targeting 2.1 -- it apparently
doesn't like seeing the 3.0 reference.

### DIFF
--- a/src/cijobs/cijobs.csproj
+++ b/src/cijobs/cijobs.csproj
@@ -2,16 +2,27 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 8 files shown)


BruceForstall:
@sandreenko Note that I don't plan to merge this until 3.0 ships, and we make everyone use it for jitutils. Or, someone figures out how to build jitutils with either 2.2 or 3.0.

Note also that jitutils depends on System.CommandLine, which is "archived" and no longer supported :-(

PR #182, opened 2018-11-02
Fix issues with DRIVEALL and convert jit-dasm-pmi to use it

BruceForstall:


src/pmi/pmi.cs
```diff
@@ -339,33 +351,27 @@ public PrepareAll(int firstMethod, bool verbose, bool time) : base(firstMethod,
...
-        if (_verbose)
-        {
-            WriteAndFlushNextMethodToPrepMarker();
-        }
+        // Many concurrent PMIs are invoked by jit-diff / jit-dasm-pmi.
+        // We need to make sure the file names we choose are unique
+        // as well as being predictable by DRIVEALL. We construct a
+        // filename based on the full pathname of the assembly being
+        // used.
+        WriteAndFlushNextMethodToPrepMarker();
```
BruceForstall: The rest of this code is to use a munged full path as the marker filename for an assembly. I presume this should be unique enough to satisfy the concern here.

AndyAyersMS: Seems like it should be good enough, yeah.

Issue #186, opened 2019-01-07
jitutils are broken on ubuntu with 3.0


sandreenko:
So I tried to clone a clean repo and ran bootstrap.sh. It showed many errors and did not build most tools, i.e.:
```
/home/sergey/git/tools2/jitutils/src/jit-diff/jit-diff.csproj : error NU1605: Detected package downgrade: System.IO.FileSystem.Primitives from 4.3.0 to 4.0.1. Reference the package directly from the project to select a different version. 
/home/sergey/git/tools2/jitutils/src/jit-diff/jit-diff.csproj : error NU1605:  jit-diff -> Microsoft.DotNet.Cli.Utils 1.0.1 -> NuGet.ProjectModel 4.0.0-rtm-2283 -> NuGet.DependencyResolver.Core 4.0.0-rtm-2283 -> NuGet.Protocol.Core.v3 4.0.0-rtm-2283 -> NuGet.Protocol.Core.Types 4.0.0-rtm-2283 -> NuGet.Configuration 4.0.0-rtm-2283 -> System.Xml.XDocument 4.0.11 -> System.Xml.ReaderWriter 4.0.11 -> System.IO.FileSystem 4.0.1 -> runtime.unix.System.IO.FileSystem 4.3.0 -> System.IO.FileSystem.Primitives (>= 4.3.0) 
/home/sergey/git/tools2/jitutils/src/jit-diff/jit-diff.csproj : error NU1605:  jit-diff -> Microsoft.DotNet.Cli.Utils 1.0.1 -> NuGet.ProjectModel 4.0.0-rtm-2283 -> NuGet.DependencyResolver.Core 4.0.0-rtm-2283 -> NuGet.Protocol.Core.v3 4.0.0-rtm-2283 -> NuGet.Protocol.Core.Types 4.0.0-rtm-2283 -> NuGet.Configuration 4.0.0-rtm-2283 -> System.Xml.XDocument 4.0.11 -> System.Xml.ReaderWriter 4.0.11 -> System.IO.FileSystem.Primitives (>= 4.0.1)
/home/sergey/git/tools2/jitutils/src/jit-diff/jit-diff.csproj : error NU1605: Detected package downgrade: System.Runtime.Handles from 4.3.0 to 4.0.1. Reference the package directly from the project to select a different version. 
[...]

PR #188, opened 2019-01-15
Use net core app3 if available

sandreenko:
This change is based on #181 and adds these commits:

18d9a56: Use netcoreapp3.0 if available.

Allows to run [dotnet/performanc](https://github.com/dotnet/performancee) and arm64 (they both require 3.0 to be installed).

13050c6: Force the new project system for *.cs projects.

Otherwise it does not see `TargetFramework` from the props file and uses the old one that doesn't show *.cs files in VS (because that are not included by default in the old version), but it also doesn't allow to include them because then you will hit build errors (one file is included twice).
See the issue description dotnet/project-system#1358 and the workaround is at the end of the thread.

e5f39ab: Separate common.props into two files.

One defines `<TargetFramework>`, the other `<PackageReference Include>`.
This change allows you to switch to another version (2.2 for example) simply by editing target-framework.props.

b735eb6: Delete `Target framework` argument for build.cmd/sh.

We do use `<TargetFrameworks>` anymore and there is no need to choose tfm from these scripts.

After that changes you can run jitutils on a machine with only 2.1 installed or on a machine that has only 3.0. Also, it works correctly when you open the solution in VS.

The last step in the build process doesn't pass on 3.0 because of dotnet/core-setup#4884, but will work once that is fixed.

Contributes to #186.

### DIFF
--- a/src/cijobs/cijobs.csproj
+++ b/src/cijobs/cijobs.csproj
@@ -2,16 +2,27 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <TargetFrameworks>netcoreapp1.0;netcoreapp2.0;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>
     <Compile Include="..\util\util.cs" Link="util.cs" />
   </ItemGroup>
 
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' != 'netcoreapp3.0' ">
     <PackageReference Include="Microsoft.DotNet.Cli.Utils" Version="1.0.1" />
+    <PackageReference Include="System.CommandLine" Version="0.1.0-e170320-1" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.0' ">
+    <PackageReference Include="Microsoft.DotNet.Cli.Utils" Version="2.0.0" />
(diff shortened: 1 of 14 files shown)


AndyAyersMS:
Does multi-targeting not work here?

Please make sure test\scripts\format.py over in coreclr still works as is, otherwise merging this will break the formatter legs we run there (in both master and perhaps in the 2.1 / 2.2 servicing branches).

src/cijobs/cijobs.csproj
```diff
@@ -1,18 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), target-framework.props))\target-framework.props" />
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), jit-include.props))\jit-include.props" />
```
RussKeldorph: Just learned about #20582 in case you'd like to use it.

sandreenko:
> Does multi-targeting not work here?

As I was told If you are targeting 3.0 (in list of TargetFrameworks), you need a 3.0 SDK to build. So it is not enough to have one installed version from <TargetFrameworks> to build the project, you need all of them.

> Please make sure test\scripts\format.py over in coreclr still works as is, otherwise merging this will break the formatter legs we run there (in both master and perhaps in the 2.1 / 2.2 servicing branches).

Done, it works as expected.

> Just learned about #20582 in case you'd like to use it.

dotnet/coreclr#20582, I have not known about this command, I will try to use it.

sandreenko:
I am going to merge it. If we find a scenario where we do not have VS 2017 and have to use an old msbuild version I will revert the last commit (Use `GetPathOfFileAbove`).

PR #190, opened 2019-01-23
Allow parsing unix git-diffs

jashook:


src/jit-analyze/jit-analyze.cs
```diff
@@ -643,18 +643,31 @@ class ScriptResolverPolicyWrapper : ICommandResolverPolicy
...
-                    if (fields.Length != 5)
+                    string manipulatedLine = line;
+
+                    int numFields = 5;
+
+                    Regex unixGitOutputRegex = new Regex(@"\{\w+\s=>\s\w+\}");
+                    if (unixGitOutputRegex.Matches(line).Count > 0)
+                    {
+                        string[] splitLine = unixGitOutputRegex.Split(line);
+                        manipulatedLine = String.Join("base", splitLine);
```
BruceForstall: I'm confused about what these two lines do. It would be helpful to write comments with an example, as well as adding an example for the case that is not in the "unix" format

PR #191, opened 2019-01-31
Fix some ifdefs so pmi works with 3.0

AndyAyersMS:
NETCOREAPP2_1 does not imply NETCOREAPP3_0. We can clean all this up
once NETSTANDARD2_1 becomes a thing.

### DIFF
--- a/src/pmi/PMIDriver.cs
+++ b/src/pmi/PMIDriver.cs
@@ -167,7 +167,7 @@ namespace PMIDriver
                 p.StartInfo.RedirectStandardError = true;
 
                 // Fetch our command line. Split off the arguments.
-#if NETCOREAPP2_1
+#if NETCOREAPP2_1 || NETCOREAPP2_2 || NETCOREAPP3_0
                 // For .Net Core the PMI assembly is an argument to dotnet.
                 string newCommandLine = Environment.CommandLine.Replace("DRIVEALL", "PREPALL");
 #else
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@sandreenko PTAL
cc @dotnet/jit-contrib

src/pmi/PMIDriver.cs
```diff
@@ -167,7 +167,7 @@ private static PrepAllResult Compute(PrepAllInfo pi)
                 p.StartInfo.RedirectStandardError = true;
 
                 // Fetch our command line. Split off the arguments.
-#if NETCOREAPP2_1
+#if NETCOREAPP2_1 || NETCOREAPP2_2 || NETCOREAPP3_0
                 // For .Net Core the PMI assembly is an argument to dotnet.
                 string newCommandLine = Environment.CommandLine.Replace("DRIVEALL", "PREPALL");
 #else
```
RussKeldorph: It's kind of surprising if this is the best way to distinguish full from core.  Would it be better to add the known full fx symbols on an #elif here so you can #error for next time?

RussKeldorph: (Just saw the comment about NETSTANDARD2_1, which mitigates my comment above somewhat)

AndyAyersMS: I don't know if it's the best way. Maybe we should ask around.

It is more annoying to have to keep stringing together those netcoreapps.

PR #196, opened 2019-02-22
Add filter option to jit-analyze

AndyAyersMS:
When analyzing a large collection of diff assemblies, you can use this to focus
analysis on a subset.


AndyAyersMS:
cc @dotnet/jit-contrib @benaadams

Sample usage:
```
jit-analyze --base D:\repos\coreclr\bin\diffs\dasmset_4\base --diff D:\repos\coreclr\bin\diffs\dasmset_4\diff --filter *AspNet* --count 10
Found 88 files with textual diffs.

Summary:  (using filter '*AspNet*')
```

benaadams:
Lovely :)

Issue #198, opened 2019-03-08
PMI: consider adding more types to use in instantiations


AndyAyersMS:
Couple of things that would make PMI more useful for catching issues:

* We should be trying both 8 bit and 16 bit short types. Perhaps one of these unsigned. Adding one of each increases the number of methods PMI'd in a checked arm64 corelib from ~26000 to ~31000.
* Would be nice if PMI could figure out how to instantiate they key methods in low-level async processing, eg `AwaitUnsafeOnCompleted`. This would probably require some special purpose pattern matching. When PMI encounters a generic with constraints it has no idea how to create types that will satisfy those constraints.

briansull:
@AndyAyersMS 

To me it appears that #199 completely address this issue.
It adds instantiations for byte and short, an 8-bit and a 16-bit sign type.
This PR also adds AwaitUnsafeOnCompleted support  for a couple of special purpose instantiations.

briansull:
I propose that we close this issue.

PR #199, opened 2019-03-15
PMI: instantiate short int types and AwaitUnsafeOnCompleted

AndyAyersMS:
Following up on #198 -- add byte and short to the instantation type lists,
and add special handling so PMI can instantiate AwaitUnsafeOnCompleted.


AndyAyersMS:
cc @dotnet/jit-contrib 

Increases "coverage" somewhat for generic types and methods.

src/pmi/pmi.cs
```diff
@@ -21,6 +22,23 @@ public static string MapPathToFileName(string path)
     }
 }
 
+class AsyncHelper
```
BruceForstall: It would be useful to have a header comment here explaining what these special-case types are for.

src/pmi/pmi.cs
```diff
@@ -988,6 +1006,29 @@ private List<MethodBase> GetInstances(Type type, MethodBase method)
             }
         }
 
+        // Special case a core async method
```
BruceForstall: It would perhaps be helpful to expound upon this comment to say why these are particularly interesting to special case..

Issue #189, opened 2019-01-18
Include ASP.NET Core assemblies in the standard set


benaadams:
When generating the wider set of jit diffs to check; probably should be recentish e.g. from 3.0 [core-sdk](https://github.com/dotnet/core-sdk)

AndyAyersMS:
If you download manually and extract it kind of works... but for some reason the cross-assembly resolution

```
jit-diff diff --base --base_root d:\repos\coreclr2 --diff --pmi --cctors --assembly C:\Users\andya\Downloads\3.0.0-preview-18579-0056
...
Dasm commands returned 13 base failures, 13 diff failures.
Completed PMI Diffs for C:\Users\andya\Downloads\3.0.0-preview-18579-0056 [invoking .cctors] in 223.78s
Diffs (if any) can be viewed by comparing: D:\repos\coreclr\bin\diffs\dasmset_5\base D:\repos\coreclr\bin\diffs\dasmset_5\diff
Analyzing diffs...
PMI Diffs for C:\Users\andya\Downloads\3.0.0-preview-18579-0056 [invoking .cctors] for x64 default jit
Summary:
(Lower is better)
Total bytes of diff: 0 (0.00% of base)
0 total files with size differences (0 improved, 0 regressed), 138 unchanged.
0 total methods with size differences (0 improved, 0 regressed), 28069 unchanged.
Completed analysis in 3.44s

Warning: Failures detected generating asm: 13 base, 13 diff

D:\repos\coreclr>jit-analyze --base D:\repos\coreclr\bin\diffs\dasmset_5\base --diff D:\repos\coreclr\bin\diffs\dasmset_5\diff

Summary:
(Lower is better)

Total bytes of diff: 0 (0.00% of base)

0 total files with size differences (0 improved, 0 regressed), 138 unchanged.

0 total methods with size differences (0 improved, 0 regressed), 28069 unchanged.
```
Typical failure case:
```
ReflectionTypeLoadException Microsoft.AspNetCore.SignalR.Common
ReflectionTypeLoadException Microsoft.AspNetCore.SignalR.Common   ex: Could not load file or assembly 'Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed'. Could not find or load a specific file. (Exception from HRESULT: 0x80131621)
ReflectionTypeLoadException Microsoft.AspNetCore.SignalR.Common type: JsonUtils
[...]

PR #202, opened 2019-03-22
jit-dasm-pmi: set PMIPATH properly

### LINKED ISSUE #201: jit-dasm-pmi: PMIPATH not set properly
The PMIPATH ends up empty here:

https://github.com/dotnet/jitutils/blob/60d8f62538ee1b372e3d68671a19527a675e654a/src/jit-dasm-pmi/jit-dasm-pmi.cs#L530-L531

need to use `assemblyPath` or similar.

AndyAyersMS:
Currently PMIPATH is always set to an empty string. It should be set to the
directory of the target assembly.

Closes #201.

### DIFF
--- a/src/jit-dasm-pmi/jit-dasm-pmi.cs
+++ b/src/jit-dasm-pmi/jit-dasm-pmi.cs
@@ -528,7 +528,7 @@ namespace ManagedCodeGen
                     }
 
                     // Set up PMI path...
-                    AddEnvironmentVariable("PMIPATH", Path.GetDirectoryName(assembly.Name));
+                    AddEnvironmentVariable("PMIPATH", assembly.Path);
 
                     if (this.verbose)
                     {


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib

BruceForstall:
@AndyAyersMS Do you see any obvious improvement to PMI results? E.g., assemblies that now work when PMI'ed?

AndyAyersMS:
I have examples, yes -- if you are pointing PMI at a published directory of some dotnet project, and that directory has multiple dependent DLLs, they can now be resolved.

It may also fix the issues I saw with reach expanding proposals like #189 -- need to double check.

Bigger context is I am trying to use a non-default AssemblyLoadContext for the PMI target assemblies and their dependencies, so that if that project refers to different versions of already loaded DLLs we can still PMI them, but that part is not ready yet.

PR #204, opened 2019-05-02
Add new NuGet package source

BruceForstall:
Linux/arm64 PMI asm diffs (using 3.0 dotnet CLI) would fail to
restore without this.


BruceForstall:
fyi @dotnet/jit-contrib 
Thanks @sbomer for the fix.

PR #207, opened 2019-05-18
Have PMI use a custom assembly load context

AndyAyersMS:
Allows us flexibility to try and locate dependent assemblies, and also
avoid conflicts with the assemblies PMI itself is using.


AndyAyersMS:
Work in support of #189 -- you should now be able to point PMI at directory of inter-dependent assemblies (like asp.net) and have it resolve cross-assembly types and also avoid version conflicts.

cc @dotnet/jit-contrib @benaadams

PR #208, opened 2019-05-21
[WIP] Add publishing step

echesakovMSFT:


echesakovMSFT:
@dotnet/jit-contrib PTAL

echesakovMSFT:
I am not sure whether we need to publish for each arch/os configuration or it would be good enough to do for **any** RID

sbomer:
If you're only building managed code, I think the easiest thing to do is to publish framework-dependent, without passing a RID. That should be usable on any platform (as long as you already have a framework on the target machine - if I understand your scenario, it sounds like you'll be placing the built binaries into CORE_ROOT, which should work AFAIK). Just note that by default "dotnet publish" will still produce an apphost for the host platform - you just wouldn't use that.

echesakovMSFT:
Not going to work on this in near future.

PR #210, opened 2019-05-31
Fix PMI for 2.x SDKs

AndyAyersMS:
Tweak the `CustomLoadContext` we define for pre-netcoreapp 3.0 builds.


AndyAyersMS:
@CarolEidt this should fix the issue you it.
cc @dotnet/jit-contrib

PR #214, opened 2019-08-21
Warning base zero size in jit-diff.

sandreenko:
It is pretty easy to run diffs with a release version and get results that look valid: `Total bytes of diff: 0`.

Add a warning message to avoid that.


sandreenko:
PTAL @jashook @BruceForstall

PR #220, opened 2019-10-30
jit-format fixes

erozenfeld:
rewriteCompileCommands had two bugs that were canceling each other
with VS2017: extracting cl.exe path was incorrect when it contained
spaces and constructing a new string from options could miss a space
after /FP<path> option.

VS2017 used long file names so we ended up with an extra quote after
cl.exe and that was compensating for a lack of a space after /FP<file>
option.

VS2019 uses 8.3 path for cl.exe so the compensating quote is not there
and clang-tidy complains.

This change fixes both bugs and also cleans up a couple of places:
I collapsed identical if and else code and added checking for "Error"
in addition to "error:" in command output. I noticed that clang-tidy may
fail with just "Error" in the output.

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -459,14 +459,7 @@ namespace ManagedCodeGen
 
                 foreach (string project in config.Projects)
                 {
-                    if (config.IsWindows && config.RewriteCompileCommands)
-                    {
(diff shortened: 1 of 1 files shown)


erozenfeld:
@dotnet/jit-contrib PTAL

erozenfeld:
This unblocks https://github.com/dotnet/coreclr/pull/27514

PR #221, opened 2019-11-01
More fixes for jit-format.

erozenfeld:
1. Don't add the precompiled header option.
2. Pass -D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH so that older versions of clang-tidy.exe work in VS2019 environment.


erozenfeld:
@sandreenko @dotnet/jit-contrib PTAL

src/jit-format/jit-format.cs
```diff
@@ -555,7 +556,7 @@ public static string rewriteCompileCommands (string compileCommandFile, string p
                             {
                                 compileCommand = compileCommand + " " + option;
                             }
-                            else if (option.Contains("src/jit"))
+                            else if (option.Contains("src/jit") && !option.StartsWith("/Fp"))
```
sandreenko: Could you please add a comment about why we are ingoring "/Fp" here so it won't confuse us in the future?

erozenfeld: I added a comment. Thank you for the review.

PR #225, opened 2019-11-13
Fix header for diff tsv file

CarolEidt:
Ensure a tab before each "Base" metric,
and add a space before metric name to make it easier to view in excel (with wrap text)


CarolEidt:
@dotnet/jit-contrib PTAL

PR #226, opened 2019-11-14
Updates to jit-diff and jit-analyze

briansull:
1.  Added -count N arg support to jit-diff
2.  Updated jit-diff  help message to print the available metrics.
3.  Added Percentage of Metric columns for the TSV files in jit-analyze

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -878,7 +878,7 @@ namespace ManagedCodeGen
                     outputStreamWriter.Write($"File\tMethod");
                     foreach (Metric metric in MetricCollection.AllMetrics)
                     {
-                        outputStreamWriter.Write($"\tBase {metric.Name}\tDiff {metric.Name}\tDelta {metric.Name}");
+                        outputStreamWriter.Write($"\tBase {metric.Name}\tDiff {metric.Name}\tDelta {metric.Name}\tPercentage {metric.Name}");
                     }
                     outputStreamWriter.WriteLine();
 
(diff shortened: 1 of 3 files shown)


briansull:
```
    -m, --metric [arg]     Metric to use for diff computations.
                           Available metrics: CodeSize(default),
                           PerfScore, PrologSize
    --count <arg>          provide the count parameter to jit-analyze
                           (default 20)
```

briansull:
@AndyAyersMS PTAL

briansull:
@dotnet/jit-contrib PTAL

src/jit-analyze/jit-analyze.cs
```diff
@@ -878,7 +878,7 @@ public static void GenerateTSV(IEnumerable<FileDelta> compareList, string path)
                     outputStreamWriter.Write($"File\tMethod");
                     foreach (Metric metric in MetricCollection.AllMetrics)
                     {
-                        outputStreamWriter.Write($"\tBase {metric.Name}\tDiff {metric.Name}\tDelta {metric.Name}");
+                        outputStreamWriter.Write($"\tBase {metric.Name}\tDiff {metric.Name}\tDelta {metric.Name}\tPercentage {metric.Name}");
```
CarolEidt: Minor suggestion - could this be `\t{metric.Name} %` so it's more compact?

src/jit-diff/jit-diff.cs
```diff
@@ -178,7 +179,7 @@ public Config(string[] args)
                     syntax.DefineOption("t|tag", ref _tag, "Name of root in output directory. Allows for many sets of output.");
                     syntax.DefineOption("c|corelib", ref _corelib, "Diff System.Private.CoreLib.dll.");
                     syntax.DefineOption("f|frameworks", ref _frameworks, "Diff frameworks.");
-                    syntax.DefineOption("m|metric", ref _metric, false, "Metric to use for diff computations (default: CodeSize).");
+                    syntax.DefineOption("m|metric", ref _metric, false, "Metric to use for diff computations. Available metrics: CodeSize(default), PerfScore, PrologSize");
```
BruceForstall: Can you specify multiple metrics? If so, how?

src/jit-diff/jit-diff.cs
```diff
@@ -197,6 +198,9 @@ public Config(string[] args)
                     syntax.DefineOption("tsv", ref _tsv, "Dump analysis data to diffs.tsv in output directory.");
                     syntax.DefineOption("tier0", ref _tier0, "Diff tier0 codegen where possible.");
 
+                    // used by jit-analyze 
+                    syntax.DefineOption("count", ref _count, "provide the count parameter to jit-analyze (default 20)");
```
BruceForstall: nit: capitalize "provide"?

src/jit-analyze/jit-analyze.cs
briansull: OK

src/jit-diff/jit-diff.cs
briansull: No Andy changes only support a single metic

src/jit-diff/jit-diff.cs
briansull: OK

src/jit-diff/jit-diff.cs
briansull: And it is case sensitive

AndyAyersMS: It would not be too hard. Jit-diff could invoke jit-analyze once for each metric, or we could update jit-analyze to allow multiple metrics too.

PR #227, opened 2019-11-14
Use /artifacts in place of /bin

jashook:
@dotnet/jit-contrib ptal

This change is updating format job to use `/artifacts` in place of `/bin`. https://github.com/dotnet/runtime/pull/51 for the change


jashook:
Yes release/3.0 and release/3.1 run formatting. Those jobs should already be broken now. I would not suggest supporting both. We can in the service branches pull at an older commit.

jashook:
There are a bunch of script changes that will need to be done to adhere to the new repo layout. I am going to merge this to unblock work in dotnet/runtime. See https://github.com/dotnet/runtime/pull/51/

PR #228, opened 2019-11-21
Add support for crossgen2 to jit-dasm and jit-diff.

erozenfeld:
--crossgen arg can now point to crossgen.exe or crossgen2.exe.

crossgen.exe and crossgen2.exe have diffferent command line args
so jit-dasm takes care of passing the right args depending on whether
--crossgen points to crossgen.exe or crossgen2.exe.

### DIFF
--- a/src/jit-dasm/jit-dasm.cs
+++ b/src/jit-dasm/jit-dasm.cs
@@ -32,6 +32,12 @@ using Microsoft.DotNet.Tools.Common;
 
 namespace ManagedCodeGen
 {
+    public enum CodeGenerator
+    {
+        Crossgen,
+        Crossgen2
+    }
+
     // Define options to be parsed 
     public class Config
     {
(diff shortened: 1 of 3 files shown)


erozenfeld:
@dotnet/jit-contrib PTAL

src/jit-dasm/jit-dasm.cs
```diff
@@ -109,6 +116,20 @@ private void Validate()
...
+                    if (fullCrossgenPathLower.EndsWith("crossgen.exe"))
+                    {
+                        _codeGenerator = CodeGenerator.Crossgen;
+                    }
+                    else if (fullCrossgenPathLower.EndsWith("crossgen2.exe"))
+                    {
+                        _codeGenerator = CodeGenerator.Crossgen2;
+                    }
+                    else
+                    {
```
echesakovMSFT: This wouldn't work on Linux/macOS

src/jit-dasm/jit-dasm.cs
```diff
@@ -560,15 +588,145 @@ void AddEnvironmentVariable(string varName, string varValue)
...
+            abstract protected void AddAssembliesPathsOption(List<string> commandArgs);
+
+            abstract protected void AddOutputPathOption(List<string> commandArgs, string outputPath);
+
+            abstract protected void AddOptimizationOption(List<string> commandArgs);
+
+            abstract protected Command CreateCommand(ICommandResolverPolicy resolverPolicy, List<string> commandArgs);
+        }
+
+        private class CrossgenDisasmEngine : DisasmEngine
```
echesakovMSFT: Should the classes should be `sealed`?

src/jit-dasm/jit-dasm.cs
erozenfeld: Good catch, thanks. Fixed.

src/jit-dasm/jit-dasm.cs
erozenfeld: Added sealed.

erozenfeld:
I also changed command line list creation to use Add instead of Insert to make it linear instead of quadratic.

PR #229, opened 2019-11-21
PMI: fix race condition in timer callback

AndyAyersMS:
It's possible for timer to be null when the callback fires. I saw
this a few times when gathering `--tier0` diffs via `jit-diffs`.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -1302,8 +1302,11 @@ class PrepareMethodinator
 
     private static void DummyTimerCallback(object state)
     {
-        timer.Change(Timeout.Infinite, Timeout.Infinite);
-        flag = 1;
+        if (timer != null)
+        {
+            timer.Change(Timeout.Infinite, Timeout.Infinite);
+            flag = 1;
+        }
     }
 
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
cc @dotnet/jit-contrib

src/pmi/pmi.cs
```diff
@@ -1313,7 +1316,7 @@ private static void EnsureTimerCallbackIsJitted()
             callback: new TimerCallback(DummyTimerCallback),
             state: null,
             dueTime: 10,
-            period: Timeout.Infinite);
+            period: 1000);
```
BruceForstall: Wouldn't it be better to start it disabled, let the assignment happen, then change it, e.g.,:
```
timer = new Timer(
            callback: new TimerCallback(DummyTimerCallback),
            state: null,
            dueTime: Timeout.Infinite,
            period: Timeout.Infinite);
timer.Change(10, Timeout.Infinite);
```

then, no change needed in `DummyTimerCallback`?

AndyAyersMS: Thanks. That seems simpler.

PR #234, opened 2019-11-26
Add --optimize option to calls to crossgen2.exe.

erozenfeld:
This is a follow-up to #228. Unlike crossgen.exe, crossgen2.exe
needs an explicit --optimize option. I added the method to set it
in #228 but forgot to call it from GenerateAsm. This change adds
the missing call.


erozenfeld:
@echesakovMSFT @dotnet/jit-contrib PTAL

PR #235, opened 2019-11-26
Allow both "perf score" and "PerfScore" is the Disassembly output

briansull:
I have changed the JIT to always use the term PerfScore everyplace.
This updates the regex to match either "perf score" or "PerfScore"


briansull:
@dotnet/jit-contrib PTAL

PR #236, opened 2019-12-01
Fix --json jit-analyze option.

erozenfeld:
1. MetricCollection.metrics was omitted by Json serializer since
it's private. The fix is to apply JsonProperty atribute to it.
2. Calling JsonConvert.SerializeObject on each file info was resulting
in an invalid json file. The fix is to call it on a filtered list of
file infos.


erozenfeld:
@AndyAyersMS @dotnet/jit-contrib PTAL

PR #239, opened 2019-12-09
Bring over superpmicollect from coreclr/runtime repo

AndyAyersMS:
This is a pristine copy to use as a starting point for PMI/SPMI mashups.

### DIFF
--- a/src/superpmi/superpmicollect.cs
+++ b/src/superpmi/superpmicollect.cs
@@ -1,0 +1,767 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Diagnostics;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace SuperPMICollection
+{
+
+    public class SpmiException : Exception
+    {
+        public SpmiException() : base()
+        { }
+
+        public SpmiException(string message)
+            : base(message)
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib

sandreenko:
Is the plane to keep https://github.com/dotnet/runtime/tree/master/src/coreclr/tests/src/JIT/superpmi in both runtime and jitutils?

Did @jashook use that cs logic for his OSX collection and replays or was it a separate python script?

BruceForstall:
Can you pick up the changes I just merged as part of https://github.com/dotnet/runtime/pull/652?

Was this a pure copy (no change) of the version in the test tree?

I'm thinking that we'll want this functionality to be repackaged (eventually) as a library as part of the PMI/SuperPMI work, making this tool itself mostly a CLI on top of that library.

> Is the plan to keep https://github.com/dotnet/runtime/tree/master/src/coreclr/tests/src/JIT/superpmi in both runtime and jitutils?

IMO, yes. While it might be useful to have the runtime tests depend on jitutils packages (which currently aren't built), like it depends on coredistools, that might not be a dependency we want dotnet/runtime to have.

> Did @jashook use that cs logic for his OSX collection and replays or was it a separate python script?

A separate script: https://github.com/dotnet/runtime/blob/master/src/coreclr/scripts/superpmi.py. It was built using the same logic, though (and even copied the comments).

AndyAyersMS:
> Was this a pure copy (no change) of the version in the test tree?

Yes, best to start off that way, I thought.

PR #241, opened 2019-12-11
Update jitutils .NET Core SDK references

BruceForstall:
- Move from 3.0 to 3.1
- If 5.0 is installed, use that
- Stop building obsolete cijobs (I didn't want to delete it just yet; maybe
it could be repurposed for AzDO?)

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -50,7 +50,7 @@ REM Do as many builds as possible; don't stop on first failure (if any).
 set __ExitCode=0
 
 REM Declare the list of projects
-set projects=jit-diff jit-dasm jit-analyze jit-format cijobs pmi jit-dasm-pmi jit-decisions-analyze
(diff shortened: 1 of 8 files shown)


BruceForstall:
@dotnet/jit-contrib

BruceForstall:
@dotnet/jit-contrib Any comments?

AndyAyersMS:
Do we know why `jit-includes` can't be pared down to just the "out-of-box" assemblies?

BruceForstall:
> Do we know why jit-includes can't be pared down to just the "out-of-box" assemblies?

Well, I haven't tried to do that lately. There's System.CommandLine, of course, but I don't know about the others.

Maybe there was an interim state of 3.1 where all the right packages weren't available?

BruceForstall:
I removed all the (now?) unnecessary assembly dependencies.

PR #242, opened 2019-12-20
Fix PMI to work with .net core 3.1.

erozenfeld:


erozenfeld:
@BruceForstall @dotnet/jit-contrib PTAL

BruceForstall:
Looks like this could be simplified by using `NETFRAMEWORK` for desktop, and `NETCOREAPP` for any .NET Core specific code. And maybe reversing the conditions so newer code is `!old1 && !old2 && ...` instead of needing to add a new `|| new7` condition for each new framework.

documentation on version preprocessor symbols: https://github.com/dotnet/docs/blob/master/includes/preprocessor-symbols.md

Issue #230, opened 2019-11-22
Update jitutils for new runtime repo structure


AndyAyersMS:
A number of jitutils tools know about the layout of build artifacts and tests in the coreclr repo.

With the consolidation of coreclr into runtime, all this needs updating.

Since jitutils is not branched or officially released, it also would be nice if jitutils tools continued to work with the coreclr repo (might need them for servicing, etc).

tannergooding:
I put up https://github.com/dotnet/jitutils/pull/244 to resolve this.

Issue #243, opened 2019-12-31
Update jitutils to support the new dotnet/runtime repo


tannergooding:
jitutils doesn't understand the new directory structure of dotnet/runtime. It should be updated (possibly with a switch) so that it will "just work"

BruceForstall:
Related: https://github.com/dotnet/jitutils/issues/230

tannergooding:
Closing in favor of #230 since it was logged first.

Issue #217, opened 2019-10-17
Consider a jit-diff mode where disassembly is done in the respective base/diff repo


tannergooding:
As from https://github.com/dotnet/jitutils/issues/216, it might be useful to expose a mode where base/diff is done in the respective repos rather than by copying the base jit into the diff repo test environment.

This would help with analyzing purely managed code changes and cases where the jit interface has changed (which prevents the jits from being interchanged).

PR #244, opened 2020-01-02
Fixup jitutils to work with the dotnet/runtime repo

tannergooding:
This resolves https://github.com/dotnet/jitutils/issues/230 ~~https://github.com/dotnet/jitutils/issues/243~~

This does not provide any switch to continue working on dotnet/coreclr, as I assume there wont be any significant changes and devs needing to test changes can always use an older snapshot of jitutils (such as commit 7e6c68b913c3b0a06821ae77a184d681d3ecefc6 or prior).

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -60,7 +60,7 @@ namespace ManagedCodeGen
                     syntax.DefineOption("a|arch", ref _arch, "The architecture of the build (options: x64, x86)");
                     syntax.DefineOption("o|os", ref _os, "The operating system of the build (options: Windows, OSX, Ubuntu, Fedora, etc.)");
                     syntax.DefineOption("b|build", ref _build, "The build type of the build (options: Release, Checked, Debug)");
-                    syntax.DefineOption("c|coreclr", ref _rootPath, "Full path to base coreclr directory");
+                    syntax.DefineOption("c|coreclr", ref _rootPath, "Full path to base runtime directory");
                     syntax.DefineOption("compile-commands", ref _compileCommands, "Full path to compile_commands.json");
                     syntax.DefineOption("v|verbose", ref _verbose, "Enable verbose output.");
                     syntax.DefineOption("untidy", ref _untidy, "Do not run clang-tidy");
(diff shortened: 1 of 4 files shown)


src/jit-format/jit-format.cs
```diff
@@ -198,11 +198,11 @@ private void validate()
                 if (!Directory.Exists(_rootPath))
                 {
                     // If _rootPath doesn't exist, it is an invalid path
-                    _syntaxResult.ReportError("Invalid path to coreclr directory. Specify with --coreclr");
+                    _syntaxResult.ReportError("Invalid path to runtime directory. Specify with --coreclr");
```
tannergooding: I kept it as `--coreclr` for back-compat. We could also expose a `--runtime` option or  only expose `--runtime` and remove `--coreclr`.

We could also just rename it to `--root` or `--root-path` to make it clearer and not tie it to the repo name 😄

BruceForstall: I vote to change it to `--runtime` but also leave `--coreclr` (hidden, if possible) for back compat. `-c` is used by `src\coreclr\tests\scripts\format.py` in the CI formatting jobs.

tannergooding:
CC. @AndyAyersMS since you logged https://github.com/dotnet/jitutils/issues/230

If you feel strongly about supporting both repos, I can add in some kind of switch to support coreclr and runtime; although I think just creating a tag for the rare servicing scenario might be sufficient.

The more interesting reason for both might be doing comparisons between runtime and coreclr; but that would require a switch that works for `--base` and `--diff` (and might be more beneficial if done alongside https://github.com/dotnet/jitutils/issues/217).

BruceForstall:
I don't think we need to support the old repos (as you state, you can just check out an older version of jitutils).

AndyAyersMS:
Keeping `--coreclr` aka `-c` and introducing a new `--runtime` seems prudent.

Do the CI jobs on coreclr run jit-format (looks like no, at least for 3.1). I suppose there's limited need for back compat.

BruceForstall:
I think we disabled jit-format in coreclr servicing CI. (And if we didn't, we could do that.)

src/jit-format/jit-format.cs
tannergooding: Do we want to shortcut `--runtime` as `-r`?

BruceForstall: No strong opinion, but maybe yes if it doesn't conflict with anything else?

tannergooding: It does not conflict, I've updated it to take `-r` or `--runtime` and the help message on `c|coreclr` to indicate it exists for back-compat and to use `r|runtime` moving forward (I didn't see any option to make it hidden).

BruceForstall:
Looks great. Ship it!

Issue #237, opened 2019-12-04
PMI DriveAll fails


BruceForstall:
Using "PrepAll" works, but using "DriveAll" fails:

```
f:\gh\runtime\artifacts\tests\coreclr\Windows_NT.x64.Debug\Tests\Core_Root>pmi.exe DriveAll System.Text.Encoding.CodePages.dll
ERROR: Unknown command C:\GH\JITUTILS\BIN\PMI.DLL
...
Failed PREPALL on System.Text.Encoding.CodePages.dll
```

BruceForstall:
@AndyAyersMS @dotnet/jit-contrib

AndyAyersMS:
Think the SDK changed at some point to produce exes, so the command line munging between the parent and child PMIs needs adjusting.

AndyAyersMS:
Fixed via #245

Issue #238, opened 2019-12-04
PMI usage message omits dll path search option


BruceForstall:
When give a directory instead of an assembly to `pmi.exe PrepAll` (and `DriveAll`?), a tree search is done looking for assemblies over which to run PMI. This behavior is not described in the help message:
```
Usage:

  pmi.exe Count PATH_TO_ASSEMBLY
      Count the number of types and methods in an assembly.

  pmi.exe PrepOne PATH_TO_ASSEMBLY INDEX_OF_TARGET_METHOD
      JIT a single method, specified by a method number.

  pmi.exe PrepAll PATH_TO_ASSEMBLY [INDEX_OF_FIRST_METHOD_TO_PROCESS]
      JIT all the methods in an assembly. If INDEX_OF_FIRST_METHOD_TO_PROCESS is specified, it is the first
      method compiled, followed by all subsequent methods.

  pmi.exe DriveAll PATH_TO_ASSEMBLY
      The same as PrepAll, but is more robust. While PrepAll will stop at the first JIT assert, DriveAll will
      continue by skipping that method.

Environment variable PMIPATH is a semicolon-separated list of paths used to find dependent assemblies.

Optional suffixes will change behavior:
   -CCtors will jit and run cctors before jitting other methods
   -Quiet will suppress in-progress messages for type and method exploration
   -Time will always show elapsed times, even in -Quiet mode
```

In particular, `PATH_TO_ASSEMBLY` is described as a single assembly, without the possibility that this can be a directory in which assemblies are searched.

AndyAyersMS:
Fixed via #245

PR #245, opened 2020-01-03
Various pmi fixes

AndyAyersMS:
Get things working again with netcoreapp5_0.

* updated and generalize some of the conditional compilation symbols
* log driveall invocations
* update help text
* add handler for a common exception type

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -122,7 +122,7 @@ public class Resolver
     }
 }
 
-#if NETCOREAPP3_0 || NETCOREAPP3_1
+#if NETCOREAPP3_0 || NETCOREAPP3_1 || NETCOREAPP5_0
 public class CustomLoadContext : AssemblyLoadContext
 {
     public string PmiPath { get; }
@@ -497,6 +497,11 @@ abstract class PrepareBase : CounterBase
         {
             Console.WriteLine();
             Console.WriteLine($"InvalidProgramException {type.FullName}::{method.Name}");
+        }
+        catch (System.InvalidOperationException)
+        {
+            Console.WriteLine();
+            Console.WriteLine($"InvalidOperationException {type.FullName}::{method.Name}");
         }
         catch (Exception e)
         {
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
Should fix #237. Will update to also take care of #238.

cc @dotnet/jit-contrib

src/pmi/pmi.cs
```diff
@@ -122,7 +122,7 @@ public static Assembly Resolve(string assemblyName, AssemblyLoadContext context)
     }
 }
 
-#if NETCOREAPP3_0 || NETCOREAPP3_1
+#if NETCOREAPP3_0 || NETCOREAPP3_1 || NETCOREAPP5_0
```
BruceForstall: Couldn't we reverse this, so the netcoreapp2.2 (?) and below (in the `else` clause) get the `#if` and 3.0+ get the `#else` and thus don't need to get updated going forward?

src/pmi/pmi.cs
```diff
@@ -1270,28 +1275,30 @@ private static int Usage()
...
-            + "      The same as PrepAll, but is more robust. While PrepAll will stop at the first JIT assert, DriveAll will\r\n"
-            + "      continue by skipping that method.\r\n"
-            + "\r\n"
-            + "Environment variable PMIPATH is a semicolon-separated list of paths used to find dependent assemblies.\r\n"
-            + "\r\n"
-            + "Optional suffixes will change behavior:\r\n"
-            + "   -CCtors will jit and run cctors before jitting other methods\r\n"
-            + "   -Quiet will suppress in-progress messages for type and method exploration\r\n"
-            + "   -Time will always show elapsed times, even in -Quiet mode"
+            "Usage:\n"
```
BruceForstall: I presume Console.WriteLine does embedded `\n` => `\r\n` translation on Windows? (but not on Unix)

src/pmi/pmi.cs
AndyAyersMS: Sure.

src/pmi/pmi.cs
AndyAyersMS: I can't find anything official, but it works fine; I've never actually used explicit `\r\n`.

BruceForstall: I presume explicit `\r\n` isn't great on Unix.

PR #248, opened 2020-01-14
PMI: add another async special case

AndyAyersMS:
Update PMI to jit the method from dotnet/runtime#1713.


AndyAyersMS:
cc @dotnet/jit-contrib

PR #249, opened 2020-01-23
Revert jit-format to using --coreclr, not --runtime

BruceForstall:
The change to `--runtime` pointing to the runtime repo root
was not fully plumbed through, so if you didn't pass it explicitly
the src\coreclr directory, jit-format didn't work. (Note that the
CI jobs do pass this explicitly.)

Remove the `--runtime` argument, and verify that the `--coreclr`
directory is correct.

Added some additional logging in `verbose` mode about what command
lines are being executed.


BruceForstall:
@dotnet/jit-contrib PTAL

PR #251, opened 2020-02-15
Don't hard-code framework assemblies.

erozenfeld:
With this change we are enumerating assemblies in core_root and
then sort them by size (with SPC always first).

We don't filter out assemblies with no methods (e.g.,
forwarder-only assemblies) upfront but that doesn't seem to matter.


erozenfeld:
@dotnet/jit-contrib

src/jit-diff/diff.cs
```diff
@@ -390,6 +390,39 @@ public static int DiffCommand(Config config)
...
+
+                string assembly2 = assemblyInfo2.Path;
+
+                if (assembly2 == s_CoreLibAssembly)
+                {
+                    return 1;
+                }
+
+                long length1 = new FileInfo(assembly1).Length;
+                long length2 = new FileInfo(assembly2).Length;
```
briansull: Calling this in the sort compare function will create a large number of unused FileInfo objects.

erozenfeld: They will be used, just very briefly, and then collected. In our scenario the sort happens instantaneously, I don't think this is worth optimizing.

erozenfeld:
I pushed a change to exclude `xunit.performance.api.dll` since it has a reference to non-existing `xunit.runner.utility.dll`, which caused pmi errors. This is temporary until we remove obsolete xunit.performance from the build. @AndyAyersMS PTAL.

erozenfeld:
macOS failure is unrelated to this change.

PR #252, opened 2020-02-26
Fix PMI non-determinism and a potential deadlock.

erozenfeld:
PMI in DRIVEALL mode invokes subprocesses in PREPALL mode
in order to be able to recover from errors. We had COMPlus environment
variables (including COMPlus_JitStdOutFile) set for both the parent and
child processes so the jit emitted disassembly, etc. to the same file.
The parent process was still active when the child was running (e.g.,
it called p.StandardOutput.ReadToEnd()). That resulted in duplicate
methods (especially from System.Private.Corelib) in the dasm file and
non-deterministic output interleaving.

The fix is not to set the COMPlus environment variables  in jit-dasm-pmi
when invoking pmi but instead pass them in PMIENV environment variable
as semicolon-separated name=value pairs and set the actual COMPlus
environment variables only on child processes.

I also fixed a potential deadlock in output/error redirection: at least
one of the redirected steams has to be read asynchronously.
See https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput
for details.

### DIFF
--- a/src/jit-dasm-pmi/jit-dasm-pmi.cs
+++ b/src/jit-dasm-pmi/jit-dasm-pmi.cs
@@ -20,6 +20,7 @@ using System.IO;
 using System.Collections.Generic;
 using System.Reflection;
 using System.Linq;
+using System.Text;
 using Microsoft.DotNet.Cli.Utils;
 using Microsoft.DotNet.Tools.Common;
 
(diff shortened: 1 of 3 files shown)


erozenfeld:
@dotnet/jit-contrib @AndyAyersMS PTAL

PR #253, opened 2020-02-26
One more PMI determinism fix.

erozenfeld:
ArrayPool<> is used in several places during the pmi execution.
In particular, ArrayPool<ReadOnlyMemory<>>.Shared.Return is called.
That method registers Gen2GcCallbackFunc on Gen2GcCallback object.
Gen2GcCallbackFunc is then called from Gen2GcCallback finalizer after
a gc. At that point Gen2GcCallbackFunc is jitted. Since gc time is not
deterministic, the disasm output for Gen2GcCallbackFunc and methods
it calls was not deterministic.

The fix is to force the registration of Gen2GcCallbackFunc and then
force a gc so that Gen2GcCallbackFunc is jitted.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -14,6 +14,7 @@ using System.Runtime.Loader;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Buffers;
 
 class Util
 {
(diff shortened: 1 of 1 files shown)


erozenfeld:
@dotnet/jit-contrib @AndyAyersMS PTAL

PR #255, opened 2020-03-20
Determinism changes.

erozenfeld:
1. The timer workaround is no longer needed since EventPipe file polling
was removed in https://github.com/dotnet/coreclr/pull/24225

2. https://github.com/dotnet/runtime/pull/467 introduced a change that
causes pmi non-determinism.
System.Threading.Thread.s_isProcessorNumberReallyFast can have different
values on two invocations of the process on the same machine.
(https://github.com/dotnet/runtime/blob/aedf8f52006619ef5d4eca65d79f42cc4b7bc402/src/coreclr/src/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs#L502)
That causes non-determinism in generated code in methods
inlining System.Threading.Thread.GetCurrentProcessorId()
(https://github.com/dotnet/runtime/blob/aedf8f52006619ef5d4eca65d79f42cc4b7bc402/src/coreclr/src/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs#L492-L498)
The workaround is to set the value of
System.Threading.Thread.s_isProcessorNumberReallyFast to true via
reflection.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -12,7 +12,6 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.Loader;
 using System.Text.RegularExpressions;
-using System.Threading;
 using System.Threading.Tasks;
 using System.Buffers;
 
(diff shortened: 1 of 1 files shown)


erozenfeld:
@AndyAyersMS @sandreenko @dotnet/jit-contrib PTAL

PR #259, opened 2020-04-08
Fix the loop calling clang-tidy.

erozenfeld:
`formatOk` wasn't updated in a thread-safe way, which resulted in errors
occasionally not being reported from formatting jobs.

Fixes https://github.com/dotnet/runtime/issues/34336

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -615,7 +615,10 @@ namespace ManagedCodeGen
             {
                 Parallel.ForEach(filenames, (filename) =>
                     {
-                        formatOk &= DoClangTidyInnerLoop(fix, ignoreErrors, checks, compileCommands, filename, verbose);
+                        if (!DoClangTidyInnerLoop(fix, ignoreErrors, checks, compileCommands, filename, verbose))
+                        {
+                            formatOk = false;
+                        }
                     });
             }
 


erozenfeld:
@dotnet/jit-contrib PTAL

erozenfeld:
Does anyone know what should be done about this error: https://dev.azure.com/dnceng/public/_build/results?buildId=592432&view=logs&j=d71c4c56-574b-5ba9-9268-feb3539ccbff
```
##[warning]An image label with the label macOS-10.13 does not exist.
,##[error]The remote provider was unable to process the request.
Pool: Azure Pipelines
Image: macOS-10.13
Started: Today at 5:36 PM
Duration: 6m 48s
```

BruceForstall:
Looks like Azure Pipelines removed support for macOS 10.13: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops

We should change `azure-pipelines.yml`:
```
vmImage: macOS-10.13
```
to:
```
vmImage: macOS-10.15
```
or, perhaps better:
```
vmImage: macOS-latest
```

erozenfeld:
Thank you @BruceForstall . I pushed an update to `azure-pipelines.yml` . Now we get
```
clang-format not found here: https://clrjit.blob.core.windows.net/clang-tools/osx.10.15-x64/clang-format
clang-tidy not found here: https://clrjit.blob.core.windows.net/clang-tools/osx.10.15-x64/clang-tidy
```

@jashook How can we upload clang-format and clang-tidy to these urls?

echesakovMSFT:
> I pushed an update to azure-pipelines.yml . Now we get
> clang-format not found here: https://clrjit.blob.core.windows.net/clang-tools/osx.10.15-x64/clang-format
> clang-tidy not found here: https://clrjit.blob.core.windows.net/clang-tools/osx.10.15-x64/clang-tidy
> 
> How can we upload clang-format and clang-tidy to these urls?

@erozenfeld As a workaround you can update bootstrap.sh so on newer macOS we always download osx.10.13 packages (or whatever latest available on the server). I think they should be runnable on latest versions of macOS even though they built targeting older one.

erozenfeld:
I reverted the `azure-pipelines.yml` change. I will deal with that outside of this PR.

PR #262, opened 2020-04-23
Update macOS version.

erozenfeld:
Azure Pipelines removed support for macOS 10.13 so we need
to update azure-pipelines.yml to macOS 10.15.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,4 +21,4 @@ jobs:
   parameters:
     agentOs: macOS
     pool:
-      vmImage: macOS-10.13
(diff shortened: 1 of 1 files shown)


erozenfeld:
@dotnet/jit-contrib

echesakovMSFT:
@erozenfeld Have you updated the published clang-tidy/clang-format package to support the newer versions of macOS?

erozenfeld:
Yes, I copied clang-tidy and clang-format to https://clrjit.blob.core.windows.net/clang-tools/osx.10.15-x64

PR #264, opened 2020-04-25
Add --nodiffable option

BruceForstall:
Add --nodiffable to jit-dasm, jit-dasm-pmi, jit-diff.
This is useful for using these tools to generate disassembly
to examine, not to diff.

Also, add some more output if "dotnet --info" doesn't work
as expected.


BruceForstall:
@dotnet/jit-contrib

PR #260, opened 2020-04-08
Make jit-format use build-runtime scripts instead of build scripts.

jkoritzinsky:
This is the last change needed to remove the build.cmd/build.sh CoreCLR scripts in dotnet/runtime.

PR #270, opened 2020-05-13
Simplify jit-format's check for the coreclr directory.

AndyAyersMS:
Don't look for build.cmd or build.sh anymore.


AndyAyersMS:
Never mind, redundant with #260.

PR #272, opened 2020-06-19
Undo a non-determinism workaround.

### LINKED ISSUE #271: jit-diff fails with "Cannot set initonly static field ..."
Running a jit-diff, such as `..\jitutils\bin\jit-diff.bat diff --diff --pmi --benchmarks`, currently fails with an error similar to the following:

```
Unhandled exception. System.FieldAccessException: Cannot set initonly static field 's_isProcessorNumberReallyFast' after type 'System.Threading.Thread' is initialized.
   at System.RuntimeFieldHandle.SetValue(RtFieldInfo field, Object obj, Object value, RuntimeType fieldType, FieldAttributes fieldAttr, RuntimeType declaringType, Boolean& domainInitialized)
   at System.Reflection.RtFieldInfo.SetValue(Object obj, Object value, BindingFlags invokeAttr, Binder binder, CultureInfo culture)
   at System.Reflection.FieldInfo.SetValue(Object obj, Object value)
   at PrepareMethodinator.EnsureGetCurrentProcessorIdIsDeterministic() in D:\tagoo\Repos\jitutils\src\pmi\pmi.cs:line 1332
   at PrepareMethodinator.Main(String[] args) in D:\tagoo\Repos\jitutils\src\pmi\pmi.cs:line 1363
```

erozenfeld:
I added a workaround for disassembly non-determinism in #255.
See #255 for details of what causes the non-determinism.
It turns out it relied on a reflection hole that allowed
init-only static fields to be modified after static constructor has
been called. That hole was fixed in https://github.com/dotnet/runtime/pull/37849
so the workaround is no longer valid and causes an exception from pmi.

I don't see a way to work around the non-determinism without changing
framework code so for now I'm just reverting the workaround. Unfortunately, that
means that we can get non-deterministic disassembly for any method that inlines
System.Threading.Thread.GetCurrentProcessorId.

Fixes #271.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -1326,15 +1326,6 @@ class PrepareMethodinator
         return 101;
     }
 
-    private static void EnsureGetCurrentProcessorIdIsDeterministic()
-    {
-#if NETCOREAPP
-        var type = typeof(System.Threading.Thread);
-        var field = type.GetField("s_isProcessorNumberReallyFast", BindingFlags.NonPublic | BindingFlags.Static);
-        field.SetValue(null, true);
-#endif
-    }
-
     private static void EnsureGen2GcCallbackFuncIsJitted()
     {
 #if NETCOREAPP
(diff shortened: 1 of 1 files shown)


erozenfeld:
@dotnet/jit-contrib PTAL

kunalspathak:
I remember seeing some error around this on ARM64 when I was doing `jit-diff.dll diff --pmi`.

PR #274, opened 2020-07-18
Fix error message in jit-format

am11:
Currently without the OS option, jit-format does not autodetects the platform but instead produces the following error:

```sh
error: Specify --arch, --plaform, and --build for clang-tidy run.
```

if we follow the instruction from error message, it gives:

```sh
# with typo
error: invalid option --plaform
# without typo
error: invalid option --platform
```
The actual option it is looking for is `--os`.
Also fixed the OS name in example for Linux portable builds, which is default flavor in `dotnet/runtime`.


BruceForstall:
Thanks for the fix.

PR #276, opened 2020-07-25
Display the generic arguments of a function in PMI messages

BruceForstall:


BruceForstall:
> Seems like there ought to be logic for this in the BCL already?

Maybe? I'm not much of a C# / BCL programmer :)

> You might want to do likewise for generic types, if they're not already showing their type params.

It looks to me like they are already displayed. But, I was mostly noticing 6 identical copies of "Preparing... Foo", which confused me and led to this.

BruceForstall:
cc @dotnet/jit-contrib

PR #280, opened 2020-09-09
Add instruction count, variable debug info metrics

### LINKED ISSUE #266: jit-analyze: add instruction count metric
Currently, we have a code size metric that is in bytes of code. Add an additional metric that is number of instructions. This can be useful when comparing cross-architecture generated code, such as x64 versus arm64, to avoid to encoding effects. This might also be useful same-architecture.

kunalspathak:
Adds "Instruction Count" metrics in `jit-analyze` and `jit-diff`

Update: Also includes Variable debug info metrics from #281

Fixes #266

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -515,83 +515,61 @@ namespace ManagedCodeGen
         public static IEnumerable<MethodInfo> ExtractMethodInfo(string filePath)
         {
             Regex namePattern = new Regex(@"for method (.*)$");
-            Regex dataPattern = new Regex(@"code ([0-9]{1,}), prolog size ([0-9]{1,})");
+            Regex codeAndPrologSizePattern = new Regex(@"code ([0-9]{1,}), prolog size ([0-9]{1,})");
             // use new regex for perf score so we can still parse older files that did not have it.
-            Regex dataPattern2 = new Regex(@"(PerfScore|perf score) (\d+(\.\d+)?)");
+            Regex perfScorePattern = new Regex(@"(PerfScore|perf score) (\d+(\.\d+)?)");
+            Regex instrCountPattern = new Regex(@"instruction count ([0-9]{1,})");
 
-            var lines = File.ReadLines(filePath);
-            int currInstrCount = 0;
-            int index = 0;
-            string _methodName = null;
-            int _functionOffset = 0;
+            var result =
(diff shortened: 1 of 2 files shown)


kunalspathak:
@dotnet/jit-contrib

AndyAyersMS:
My preferences is to have the jit compute the count and add it to the summary line alongside bytes of code.

This is how we do the other metrics and this is the model we'll likely adopt for any "big data" exploration where we can't really afford to generate or scan the textual assembly listing.

kunalspathak:
That makes sense Andy. I will see what changes are needed in jit to compute and propagate that and perhaps update this PR to parse that data.

AndyAyersMS:
Also I have some changes that I never PR'd to allow debug info metrics. Let me dig those up.

kunalspathak:
@AndyAyersMS  - I have sent out https://github.com/dotnet/runtime/pull/42040 that adds instruction count as well as variable debug info in JIT. Also updated this PR to capture that information in `jit-diff` and `jit-analyze`

AndyAyersMS:
Did you push your recent updates? I don't see them.

kunalspathak:
Ah, sorry. I pushed it now.

AndyAyersMS (approved):
Looks good.

BruceForstall (approved):

BruceForstall:
@AndyAyersMS Now that we have multiple metrics: doesn't jit-analyze compute all the metrics but only output for one? Seems like maybe "-m" should take multiple desired metrics so diff parsing only happens once. And maybe the default should be Code size and Perf Score both get displayed?

kunalspathak:
> jit-analyze compute all the metrics but only output for one

I thought about it while working on this PR. However, my personal experience is if we show more metrics, it will be hard to parse the output as a user because for any metrics, apart from summary diff, we should % methods, % assembly numbers as well. Probably another thing we can do is just summarize the results of all the metrics and then display the details of one that user passed in `- m`?

BruceForstall:
> Probably another thing we can do is just summarize the results of all the metrics and then display the details of one that user passed in - m?

I currently don't have much experience with the metrics, so I'll defer to others who do.

sandreenko:
I would prefer to have an option to dump all details for all metrics. I usually run diffs in a bat for all altjits/with "--count 10000" and just leave it for an hour. So when it finishes I have all information that I could possibly want for analysis, but this is my personal preference.

PR #281, opened 2020-09-09
Support for some simple debug metrics.

AndyAyersMS:
Requires matching change in the jit.

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -214,6 +214,26 @@ namespace ManagedCodeGen
             public override string ValueString => $"{Value:F2}";
         }
 
+        public class DebugClauseMetric : Metric
+        {
+            public override string Name => "DebugClauseCount";
+            public override string DisplayName => "Debug Clause Count";
+            public override string Unit => "Clause";
+            public override bool LowerIsBetter => true;
+            public override Metric Clone() => new DebugClauseMetric();
+            public override string ValueString => $"{Value}";
+        }
+
+        public class DebugVarMetric : Metric
+        {
+            public override string Name => "DebugVarCount";
+            public override string DisplayName => "Debug Variable Count";
+            public override string Unit => "Variable";
+            public override bool LowerIsBetter => true;
+            public override Metric Clone() => new DebugVarMetric();
+            public override string ValueString => $"{Value}";
+        }
+
         public class MetricCollection
         {
             private static Dictionary<string, int> s_metricNameToIndex;
@@ -221,7 +241,7 @@ namespace ManagedCodeGen
 
             static MetricCollection()
             {
-                s_metrics = new Metric[] { new CodeSizeMetric(), new PrologSizeMetric(), new PerfScoreMetric() };
+                s_metrics = new Metric[] { new CodeSizeMetric(), new PrologSizeMetric(), new PerfScoreMetric(), new DebugClauseMetric(), new DebugVarMetric() };
                 s_metricNameToIndex = new Dictionary<string, int>(s_metrics.Length);
 
                 for (int i = 0; i < s_metrics.Length; i++)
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@kunalspathak FYI -- still need to find the matching jit changes.
cc @dotnet/jit-contrib

AndyAyersMS:
Here are the matching jit changes. The only bit I don't like is the need to repeat the method name, as our pattern matching is "stateless". It is somewhat of a requirement as currently we need to handle interleaved disasm output.
```diff
--- a/src/coreclr/src/jit/ee_il_dll.cpp
+++ b/src/coreclr/src/jit/ee_il_dll.cpp
@@ -805,8 +805,20 @@ void Compiler::eeDispVar(ICorDebugInfo::NativeVarInfo* var)
 // Same parameters as ICorStaticInfo::setVars().
 void Compiler::eeDispVars(CORINFO_METHOD_HANDLE ftn, ULONG32 cVars, ICorDebugInfo::NativeVarInfo* vars)
 {
-    printf("*************** Variable debug info\n");
-    printf("%d live ranges\n", cVars);
+    // Estimate number of unique vars with debug info
+    ALLVARSET_TP uniqueVars(AllVarSetOps::MakeEmpty(this));
+    for (unsigned i = 0; i < cVars; i++)
+    {
+        // ignore "special vars" and out of bounds vars
+        if ((((int)vars[i].varNumber) >= 0) && (vars[i].varNumber < lclMAX_ALLSET_TRACKED))
+        {
+            AllVarSetOps::AddElemD(this, uniqueVars, vars[i].varNumber);
+        }
+    }
+
+    printf("; Variable debug info: %d live ranges, %d vars for method %s\n", cVars,
+           AllVarSetOps::Count(this, uniqueVars), info.compFullName);
+
     for (unsigned i = 0; i < cVars; i++)
     {
         eeDispVar(&vars[i]);
```
Generally speaking, I think we should modify the disasm dump in a few ways:
* introduce a "summary" mode that just prints metrics and no details
* introduce a standard header that creates a state transition in the parsers (metrics that follow are for "this method")
* each metric or metric group has a standard intro text `;@Metric` or similar
* introduce a standard footer to trigger a "done " transition in the parser
* prevent record interleaving; likely a simple lock will suffice, though we need to be careful that this gets properly released on failure. Or introduce a standard thread prefix, interlock output at the line level, and require parsers to cope with de-interlacing.

kunalspathak:
Thanks @AndyAyersMS. I can include your debug info changes in my PR that adds instruction count in JIT. We will do a brainstorming about the general "disasm dump changes" and take action based on the outcome.

BruceForstall:
@AndyAyersMS I think your ideas for improved JIT dumping merits its own issue in dotnet/runtime (and perhaps a companion issue here)

AndyAyersMS:
@BruceForstall sure, I'll open one. Some of this thinking dovetails with dotnet/runtime#10462 where we would like to foster a much richer set of metrics, and the tooling to summarize and analyze them.

Also note that we have `jit-decisions-analyze` which leverages eventing to gather stats -- my hope is that the "summary mode" stats would be compact enough that we could consider emitting them as events, and so allow metric collections from release builds.

kunalspathak:
@AndyAyersMS  - "Variable debug info" in `eeDispVars()` is only printed in JITDumps and not in JitDisasm. The metrics are based on JitDisasm. Are you suggesting to include variable debug info for JitDisasm as well?

AndyAyersMS:
It's also emitted via `opts.dspDebugInfo` which maps to `COMPlus_JitDebugDump` which is enabled via  `--debuginfo` in `jit-diff`. So if you want to get debug metrics you need to pass this arg.

Generally we probably want to emit metric info all the time and not just when asked for, provided it is not too expensive to collect and aggregate.

kunalspathak:
https://github.com/dotnet/jitutils/pull/280 contains the changes from this PR. Closing this PR.

JulieLeeMSFT:
> Also note that we have `jit-decisions-analyze` which leverages eventing to gather stats -- my hope is that the "summary mode" stats would be compact enough that we could consider emitting them as events, and so allow metric collections from release builds.

Emitting the "summary mode" from release builds will be very useful. Windows ARM perf team will be able to use those telemetry.

PR #282, opened 2020-09-10
Rename protononjit.dll to clrjit_win_arm64_x64.dll

kunalspathak:
Now that we have stopped producing `protononjit.dll` in https://github.com/dotnet/runtime/pull/41126, replace its reference with new name e.g. `clrjit_win_arm64_x64.dll`

### DIFF
--- a/doc/diffs.md
+++ b/doc/diffs.md
@@ -193,7 +193,7 @@ The "jit-diff diff" command has this help message:
         --diff_root <arg>     Path to root of diff dotnet/coreclr repo.
         --arch <arg>          Architecture to diff (x86, x64).
         --build <arg>         Build flavor to diff (Checked, Debug).
-        --altjit <arg>        If set, the name of the altjit to use (e.g., protononjit.dll).
+        --altjit <arg>        If set, the name of the altjit to use (e.g., clrjit_win_arm64_x64.dll).
         --pmi                 Generate diffs via jitting instead of running crossgen
         --assembly <arg>      Look at diffs for methods in the specified assembly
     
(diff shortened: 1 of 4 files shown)


kunalspathak:
@dotnet/jit-contrib

BruceForstall (approved):

kunalspathak:
Thanks @BruceForstall . Since the PR works, I will probably replace this PR with a one that has some relevant changes that we can test because if I merge this PR as is, it will kick off the CI runs in runtime repo which won't be useful at this point.

BruceForstall:
> if I merge this PR as is, it will kick off the CI runs in runtime repo

Why is that? The runtime repo doesn't respond to changes in jitutils.

kunalspathak:
> > if I merge this PR as is, it will kick off the CI runs in runtime repo
> 
> Why is that? The runtime repo doesn't respond to changes in jitutils.

Ah, never mind. I thought this approval was for https://github.com/BruceForstall/runtime/pull/1.

AndyAyersMS:
Just some random thoughts...

Now that the altjit naming is systematic I wonder if we should add "wildcard" support, eg run batches of diffs using all known altjit os/isa combinations or something?

Also if/when a regular jit can run as an altjit some of the file swapping done by jit-dasm-pmi can be simplified as we no longer need to replace the "real jit" -- and once we do this we can perhaps also change the concurrently model for jit-dasm-pmi to be more like the one we use for crossgen (run base and diff tasks concurrently).

kunalspathak:
> Now that the altjit naming is systematic I wonder if we should add "wildcard" support, eg run batches of diffs using all known altjit os/isa combinations or something?
> 
> Also if/when a regular jit can run as an altjit some of the file swapping done by jit-dasm-pmi can be simplified as we no longer need to replace the "real jit" -- and once we do this we can perhaps also change the concurrently model for jit-dasm-pmi to be more like the one we use for crossgen (run base and diff tasks concurrently).

I have opened https://github.com/dotnet/jitutils/issues/283 to track this.

BruceForstall:
> if/when a regular jit can run as an altjit

I think we really need to remove altjit as a compile-time concept and move it to be a run-time concept. I really don't want the "cross" jits to be built and shipped as compile-time altjits for crossgen2.

I previously opened https://github.com/dotnet/runtime/issues/41643 to track this.

Issue #149, opened 2018-06-22
jit-analyze can't parse numstat output when one of the files is missing


AndyAyersMS:
Looks like missing file is described as `/dev/null`... we should handle this case in the parser.
```
Couldn't parse --numstat output '2680   0       /dev/null => "D:\\repos\\coreclr\\bin\\diffs\\linux-fx-v3\\base/xunit.performance.core.dasm"` : 'D:\dev\null' does not exist
```

PR #284, opened 2020-09-14
Use diff-filter to filter files that are not present in both base/diff

kunalspathak:
We get a weird error if a .dasm file exist in one folder and not the other. So I decided to just use `--diff-filter` option to only diff files that are present in both folders.

I thought of updating the regex, but this one seemed more compelling. 

Fixes: https://github.com/dotnet/jitutils/issues/149


kunalspathak:
@dotnet/jit-contrib

BruceForstall (approved):

PR #285, opened 2020-09-14
Append random id to log files so they get ignored during git-diff

kunalspathak:
Today, when doing diff, we perform `git diff` on `base` and `diff`. These folders has `.dasm` as well as `.log` files. The `.log` files are comparable in size. For e.g., they are 520MB while `.dasm` files are 800MB. Once we do `git diff`, we then filter the diffs that are coming out of `.dasm` files. Thus, we unnecessarily spend time diffing the `.log` files. In https://github.com/dotnet/jitutils/pull/284, I added a way to ignore files if they are not present in both the folders. With that, I propose to add a random ID to the log files so they automatically get ignored when doing `git diff` and thus saving some time.

### DIFF
--- a/src/jit-dasm-pmi/jit-dasm-pmi.cs
+++ b/src/jit-dasm-pmi/jit-dasm-pmi.cs
@@ -359,6 +359,7 @@ namespace ManagedCodeGen
             public bool doDebugDump = false;
             public bool verbose = false;
             private int _errorCount = 0;
+            private Random _rand = new Random();
 
             public int ErrorCount { get { return _errorCount; } }
 
@@ -573,7 +574,10 @@ namespace ManagedCodeGen
                         // Generate path to the output file
                         var assemblyFileName = Path.ChangeExtension(assembly.Name, ".dasm");
                         var dasmPath = Path.Combine(_rootPath, assembly.OutputPath, assemblyFileName);
-                        var logPath = Path.ChangeExtension(dasmPath, ".log");
+
+                        // Append a random id to the log files so they don't get picked up by "git diff"
+                        string randomId = _rand.Next(0, 5000).ToString();
+                        var logPath = Path.ChangeExtension(dasmPath, $".{randomId}.log");
 
                         PathUtility.EnsureParentDirectoryExists(dasmPath);
 
--- a/src/jit-dasm/jit-dasm.cs
+++ b/src/jit-dasm/jit-dasm.cs
@@ -386,6 +386,7 @@ namespace ManagedCodeGen
             public bool verbose = false;
             private int _errorCount = 0;
             protected Dictionary<string, string> _environmentVariables;
+            protected Random _rand = new Random();
 
             public int ErrorCount { get { return _errorCount; } }
 
@@ -544,7 +545,9 @@ namespace ManagedCodeGen
 
                     if (_rootPath != null)
                     {
-                        var logPath = Path.ChangeExtension(dasmPath, ".log");
+                        // Append a random id to the log files so they don't get picked up by "git diff"
+                        string randomId = _rand.Next(0, 5000).ToString();
+                        var logPath = Path.ChangeExtension(dasmPath, $".{randomId}.log");
 
                         // Redirect stdout/stderr to log file and run command.
                         using (var outputStreamWriter = File.CreateText(logPath))


kunalspathak:
@dotnet/jit-contrib

CarolEidt (approved):
Looks like a reasonable approach to me.

BruceForstall:
Seems like kind of a hack.

Is `git diff` recursive? If not (or we can invoke it to not recurse), could we put the log files in a `log` subdirectory? E.g., base\log\xxx.log, diff\log\xxx.log?

kunalspathak:
> Seems like kind of a hack.
> 
> Is `git diff` recursive? If not (or we can invoke it to not recurse), could we put the log files in a `log` subdirectory? E.g., base\log\xxx.log, diff\log\xxx.log?

I agree. Better way is to just create separate logs folder for base and diff. With that, it will create 2 new folders `baselogs` and `difflogs` that will contain log files. So when diffing, `git diff diff base` should just work fine.
```
dasmset_62
|-- base
|-- diff
|-- baselogs
|-- difflogs
```

src/jit-dasm-pmi/jit-dasm-pmi.cs
```diff
@@ -573,9 +573,13 @@ void AppendEnvironmentVariableToPmiEnv(string varName, string varValue)
                         // Generate path to the output file
                         var assemblyFileName = Path.ChangeExtension(assembly.Name, ".dasm");
                         var dasmPath = Path.Combine(_rootPath, assembly.OutputPath, assemblyFileName);
-                        var logPath = Path.ChangeExtension(dasmPath, ".log");
+
+                        // Create logs in seperate folder so they don't get picked up by "git diff"
```
BruceForstall, line 577: nit: typo: seperate => separate

src/jit-dasm/jit-dasm.cs
```diff
@@ -512,13 +512,19 @@ public void GenerateAsm()
...
 
                     string dasmPath = null;
+                    string logPath = null;
                     if (_rootPath != null)
                     {
                         // Generate path to the output file
                         var assemblyFileName = Path.ChangeExtension(assembly.Name, ".dasm");
                         dasmPath = Path.Combine(_rootPath, assembly.OutputPath, assemblyFileName);
 
+                        // Create logs in seperate folder so they don't get picked up by "git diff"
```
BruceForstall, line 522: nit: typo: seperate => separate

BruceForstall (approved):
LGTM

Looks from the code like you've gone with something closer to my suggestion, namely:
```
dasmset_62
|-- base
     |-- logs
|-- diff
     |-- logs
```

The title of this PR is now inaccurate.

AndyAyersMS:
For more complex structures the diff / base output may be directory trees (eg if you run jit-diff on the test tree). What happens then?

You can test this by just creating a small directory tree and placing some test apps in the leaves and then using `--assembly` to point at the root of the tree.

kunalspathak:
Here is how it produces with my latest change. It is not very pleasant...Do you want me to find the common ancestor and create log folders in that?

binaries folder:
```
binaries
|--- 1
     |--- foo.dll
|--- 2
     |--- bar.dll
|--- 3
     |--- baz.dll
|--- projs.dll
```

Output:
```
dasmset_63
|--- base
     |--- projs.dasm
     |--- 1
          |--- foo.dasm
     |--- 1log
          |--- foo.log
     |--- 2
          |--- bar.dasm
     |--- 2log
          |--- bar.log
     |--- 3
          |--- baz.dasm
     |--- 3log
          |--- baz.log
    |--- projs.dasm
|---baselog
    |---projs.log
|--- diff
     |--- projs.dasm
     |--- 1
          |--- foo.dasm
     |--- 1log
          |--- foo.log
     |--- 2
          |--- bar.dasm
     |--- 2log
          |--- bar.log
     |--- 3
          |--- baz.dasm
     |--- 3log
          |--- baz.log
    |--- projs.dasm
|---difflog
    |---projs.log
```

BruceForstall:
Everything goes under a `_rootPath`, which is a user-specified argument, so you have to put everything in that path.

You could create a new `_logsRootPath` argument (e.g., "--logsOutputPath") and put everything in a tree under that. If passed the same directory as `_rootPath`, behavior would be the same as today. Or, if passed something different, the logs are separated. You would have to pass this additional argument in from `jit-diff`. It could default to the same as `_rootPath`.

PR #286, opened 2020-09-15
Fix PMI diffs for arm64 with the recent altjit renaming.

sandreenko:
Fixes jit asserts when running jit-diffs --pmi for arm64 altjit.

PTAL @dotnet/jit-contrib


briansull (approved):
Thanks for fixing this issue

PR #291, opened 2020-09-19
CoreDisTools on Arm

echesakovMSFT:
This adds support for proper recognition of movw/movt instructions pairs on Arm in `CorAsmDiff::nearDiff`.
In a case when such instructions pairs are compared, their immediates should not be compared independently. Instead their values should be used to reconstruct a 32-bit value that is loaded into a register and these reconstructed value should be compared either for identity or equality (with a user defined `Comparator`). 

I tested this approach by 
1. **superpmi collect** on on windows-arm on framework libraries
2. **superpmi replay** on windows-arm 
3. **superpmi replay** on windows-x86 (with clrjit_win_arm_x86.dll jit)

In both 2 and 3 there were no cases of false positives diffs between baseline and compared jits. (If this algorithm is disabled, almost every method becomes different according to `CorAsmDiff::nearDiff` due to some immediates - either movt or movw having different values).

This PR also specifies `LLVM_DEFAULT_TARGET_TRIPLE` to be Thumb2 not Arm and decreases number of platforms supported by each flavor of coredistools.

### DIFF
--- a/build-coredistools.cmd
+++ b/build-coredistools.cmd
@@ -8,12 +8,16 @@ set TargetOSArchitecture=%1
 
 if /i "%TargetOSArchitecture%" == "windows-arm" (
     set GeneratorPlatform=ARM
+    set LLVMHostTriple=arm-pc-windows-msvc
 ) else if /i "%TargetOSArchitecture%" == "windows-arm64" (
     set GeneratorPlatform=ARM64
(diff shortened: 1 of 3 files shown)


echesakovMSFT:
@dotnet/jit-contrib PTAL

BruceForstall (approved):
Looks great. Thanks!

PR #293, opened 2020-09-24
Add infrastructure for publishing coredistools

echesakovMSFT:
This adds support for building coredistools NuGet packages for the following *portable* RIDs:
- linux-arm
- linux-arm64
- linux-x64
- osx-x64
- win-arm
- win-arm64
- win-x64
- win-x86

I removed all the nuspec files that correspond to *non-portable* RIDs (e.g. ubuntu-14.04-x64). I also updated the `projectUrl` attribute to refer to this repository.

I added a job to Azure Pipelines that produce NuGet packages and publish them as pipeline artifacts.


echesakovMSFT:
@dotnet/jit-contrib @janvorli PTAL

src/coredistools/.nuget/Microsoft.NETCore.CoreDisTools.nuspec
```diff
@@ -2,19 +2,19 @@
...
+    <projectUrl>https://github.com/dotnet/jitutils</projectUrl>
     <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <description>Disassembly Tools for CoreCLR</description>
     <releaseNotes>Initial release</releaseNotes>
     <copyright>Copyright &#169; Microsoft Corporation</copyright>
   </metadata>
   <files>
-    <file src="runtime.json" />
+    <file src="src\coredistools\.nuget\runtime.json" />
```
janvorli, line 18: I am afraid that the resulting package will still have the problem I've been hitting with the old one - the coredistools.dll / .so not being restored into the target folder of a C# project, as there are no differences in the way the package is created and I somehow assumed that that's actually the core of the problem. The reason for my thinking this way was that in R2RDump tool, we reference two native packages - Microsoft.DiaSymReader.Native and Microsoft.NETCore.CoreDisTools. The diasym reader files get placed into the target folder of R2RDump during build, but the coredistools.dll / .so doesn't. The way we reference those packages is equivalent.
But my knowledge of how nuget packages are supposed to be constructed is zero, so maybe the problem with the old package was just that it was generated with some ancient nuget that caused that problem.

echesakovMSFT: @janvorli Let me try to do this manually with the new packages.

echesakovMSFT: @janvorli I believe this is how it works.

The native assets of Microsoft.DiaSymReader.Native is copied to the target folder because we have the following logic that is imported to R2RDump.csproj from eng\DiaSymReaderNative.targets 

```xml
  <ItemGroup Condition="'$(DotNetBuildFromSource)' != 'true'">
    <Content Include="$(NuGetPackageRoot)\microsoft.diasymreader.native\$(MicrosoftDiaSymReaderNativeVersion)\runtimes\win\native\Microsoft.DiaSymReader.Native.x86.dll">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <Visible>false</Visible>
      <Pack>false</Pack>
    </Content>
    <Content Include="$(NuGetPackageRoot)\microsoft.diasymreader.native\$(MicrosoftDiaSymReaderNativeVersion)\runtimes\win\native\Microsoft.DiaSymReader.Native.amd64.dll">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <Visible>false</Visible>
      <Pack>false</Pack>
    </Content>

    <PackageReference Include="Microsoft.DiaSymReader.Native" Version="$(MicrosoftDiaSymReaderNativeVersion)" ExcludeAssets="all"/>
  </ItemGroup>
```
while for Microsoft.NETCore.CoreDisTools the file has only this
```xml
  <ItemGroup>
    <PackageReference Include="Microsoft.NETCore.CoreDisTools">
      <Version>1.0.1-prerelease-00005</Version>
    </PackageReference>
  </ItemGroup>
```

If I add similar logic to R2RDump.csproj 
```diff
--- a/src/coreclr/src/tools/r2rdump/R2RDump.csproj
+++ b/src/coreclr/src/tools/r2rdump/R2RDump.csproj
@@ -19,8 +19,13 @@

   <ItemGroup>
     <PackageReference Include="Microsoft.NETCore.CoreDisTools">
-      <Version>1.0.1-prerelease-00005</Version>
+      <Version>1.0.1-prerelease-00006</Version>
     </PackageReference>
+    <Content Include="$(NuGetPackageRoot)\runtime.win-x64.microsoft.netcore.coredistools\1.0.1-prerelease-00006\runtimes\win-x64\native\coredistools.dll">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Visible>false</Visible>
+      <Pack>false</Pack>
+    </Content>
     <PackageReference Include="System.CommandLine">
       <Version>$(SystemCommandLineVersion)</Version>
     </PackageReference>
```
Then the target folder will be populated with coredistools.dll after I run build.cmd clr.tools -a x64 -c checked
```
D:\git\runtime\artifacts\bin\coreclr\Windows_NT.x64.Checked\R2RDump>dir /b
coredistools.dll
ILCompiler.Reflection.ReadyToRun.dll
Microsoft.DiaSymReader.Native.amd64.dll
Microsoft.DiaSymReader.Native.x86.dll
R2RDump.deps.json
R2RDump.dll
R2RDump.exe
R2RDump.runtimeconfig.dev.json
R2RDump.runtimeconfig.json
System.CommandLine.dll
```

So I don't think there is an issue with how these packages are **built** but there is definitely an issue with how they are **restored** in R2RDump.

janvorli: Hmm, so we really need to reference the runtime specific packages explicitly? I have somehow assumed it should work automatically based on the target platform. It would seem natural, but I guess I was just thinking nuget is more clever than it really is.

echesakovMSFT: We still reference only one nuget package - Microsoft.NETCore.CoreDisTools - all the others (e.g. runtime.win-x64.microsoft.netcore.coredistools) are restored automatically by nuget tool. However, we need to copy coredistools.dll from a specific location in the package that corresponds to the host environment to the target folder. I don't know if there is a way to do make nuget do this for us.

janvorli: Ok, thanks. It is not a big deal to do it this way. I just didn't want to do it in a way that would be a hack.

janvorli (approved):
LGTM, thank you!

BruceForstall (approved):
LGTM

PR #294, opened 2020-09-30
CoreDisTools on Arm64

echesakovMSFT:
The small change is to enable additional Arm64 ISAs in the dissassembler. 
Without this change, for example, the superpmi would not be able to dissassemble a method containing `AESD` instruction.


echesakovMSFT:
@dotnet/jit-contrib PTAL

AndyAyersMS (approved):

PR #295, opened 2020-10-01
Allow more supported target architectures for R2RDump

echesakovMSFT:
Change the target architectures matrix for coredistools:
- win-x86, win-x64, linux-x64, osx-x64 supports X86, ARM and AArch64.
- win-arm, win-arm64, linux-arm, linux-arm64 supports ARM and AArch64.

With this change x86/x64 R2RDump can target all the platforms with one version of coredistools binaries.

I was not sure if is useful to enable X86 target support on Arm/Arm64 hosted coredistools, so I decided not to in order to have smaller coredistools binaries on these platforms.

### DIFF
--- a/build-coredistools.cmd
+++ b/build-coredistools.cmd
@@ -2,24 +2,23 @@
 setlocal EnableDelayedExpansion EnableExtensions
 
 set TargetOSArchitecture=%1
+set LLVMTargetsToBuild=AArch64;ARM;X86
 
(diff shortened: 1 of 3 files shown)


echesakovMSFT:
@dotnet/jit-contrib PTAL

echesakovMSFT:
@dotnet/jit-contrib ping

BruceForstall (approved):

PR #298, opened 2020-11-04
One CoreDisTools NuGet package

echesakovMSFT:
Build one NuGet package that includes all the coredistools binaries.


echesakovMSFT:
@dotnet/jit-contrib PTAL

echesakovMSFT:
@janvorli This is work on consolidating all coredistools binaries into one package as we discussed a while ago. However, for R2RDump we would still need to add the appropriate MSBuild target in order to copy a library corresponding to the platform it is running on to an appropriate location (next to R2RDump executable).

echesakovMSFT:
@dotnet/jit-contrib ping

BruceForstall (changes requested):
Please remove `src/coredistools/.nuget/.Microsoft.NETCore.CoreDisTools.nuspec.swp` from the change (I presume that's some vim file).

echesakovMSFT:
> Please remove `src/coredistools/.nuget/.Microsoft.NETCore.CoreDisTools.nuspec.swp` from the change (I presume that's some vim file).

Done, thanks for catching this - it's indeed a vim file.

BruceForstall (approved):

Issue #300, opened 2020-11-17
Unable to get diff for corelib


alexcovington:
I have two builds of dotnet/runtime that I would like to compare using `jit-diff`. One is just a build off of the master branch, `runtime-master`, and the other is my own local build with some changes I've made to the JIT, `runtime`.

Whenever I try to do a diff of the two builds on `System.Private.Corelib.dll`, I get the following error:

```
PS C:\Users\acovingt\source\repos\runtime> jit-diff diff --output C:\Users\acovingt\Documents\diffs --base C:\Users\acovingt\source\repos\runtime-master\artifacts\bin\coreclr\windows.x64.Release\ --diff C:\Users\acovingt\source\repos\runtime\artifacts\bin\coreclr\windows.x64.Release\ --core_root C:\Users\acovingt\source\repos\runtime-master\artifacts\tests\coreclr\windows.x64.Release\Tests\Core_Root\ --corelib                                                                                                                                                   
Beginning Crossgen CodeSize Diffs for System.Private.CoreLib.dll                                                                                                                                                                                                                        
Unhandled exception. System.ArgumentOutOfRangeException: startIndex cannot be larger than length of string. (Parameter 'startIndex')                                                                                                                                                     
  at System.String.Substring(Int32 startIndex, Int32 length)                                                                                                                                                                                                                              
  at ManagedCodeGen.jitdiff.DiffTool.IdentifyAssemblies(String basePath, String rootPath, Config config, Boolean recursive, String searchPattern) in C:\Users\acovingt\source\repos\jitutils\src\jit-diff\diff.cs:line 532                                                              
  at ManagedCodeGen.jitdiff.DiffTool.GenerateAssemblyWorklist(Config config) in C:\Users\acovingt\source\repos\jitutils\src\jit-diff\diff.cs:line 466                                                                                                                                   
  at ManagedCodeGen.jitdiff.DiffTool.DiffCommand(Config config) in C:\Users\acovingt\source\repos\jitutils\src\jit-diff\diff.cs:line 332                                                                                                                                                
  at ManagedCodeGen.jitdiff.Main(String[] args) in C:\Users\acovingt\source\repos\jitutils\src\jit-diff\jit-diff.cs:line 1335
```

I've tried re-cloning and rebuilding both dotnet/jitutils and dotnet/runtime, but this error persists.

Here is `dotnet --info` output to help with troubleshooting:

```
PS C:\Users\acovingt\source\repos\runtime> dotnet --info
[...]

PR #287, opened 2020-09-15
Replace Dotnet.Cli.Utils.Command with System.Diagnostic.Process

kunalspathak:
- Replaced `Dotnet.Cli.Utils.Command` with `System.Diagnostic.Process` which is more robust and allow us to better handle process spawning mechanism.
- Print the `git diff` command to fix #92 
- Switch to `System.Diagnostic.Process` to fix #175.
- Add cancellation mechanism that will kill all the process and child process the given jitutil has spawned. Fixes #151 .
- Print the environment variables that fixes #141.

PR #302, opened 2020-11-19
Fix path problems so we can get diff working

kunalspathak:
- There was miscalculation happening if the `CORE_ROOT` ended with trailing `\`. Fixed it.
- Also, in https://github.com/dotnet/jitutils/pull/287, I changed the way log files were created. If there is no contents to dump, then we would skip creating the log files. However, I missed the check if we decide to move a log file. 

Fixes: #300


kunalspathak:
@dotnet/jit-contrib

BruceForstall (approved):

PR #304, opened 2020-11-24
Remove explicit "src" level from source path construction

trylek:
I believe this to be the minimum change to facilitate removal of
the duplicated "src" folder level under "src/coreclr" in the
runtime repo. Before the removal we'll need to make a coordinated
runtime change to pass the source directory as "src/jit" instead
of just "jit"; the removal PR should subsequently change the
"src/jit" folder as the input argument to jit-format to just "jit".

Thanks

Tomas

Related to: https://github.com/dotnet/runtime/pull/44973

P.S. This is my first chance at a jitutils PR, thanks in advance for your
patience with me; I currently don't know how to test this without
merging in both this change and the envisioned runtime counterpart,
please let me know if you believe there's an easier way.

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -59,6 +59,7 @@ namespace ManagedCodeGen
                     syntax.DefineOption("o|os", ref _os, "The operating system of the build (options: Windows, OSX, Linux etc.)");
                     syntax.DefineOption("b|build", ref _build, "The build type of the build (options: Release, Checked, Debug)");
                     syntax.DefineOption("c|coreclr", ref _rootPath, "Full path to base runtime/src/coreclr directory");
+                    syntax.DefineOption("s|source-directory", ref _srcDirectory, "Source directory relative to coreclr root");
                     syntax.DefineOption("compile-commands", ref _compileCommands, "Full path to compile_commands.json");
                     syntax.DefineOption("v|verbose", ref _verbose, "Enable verbose output.");
                     syntax.DefineOption("untidy", ref _untidy, "Do not run clang-tidy");
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -164,7 +164,7 @@ namespace ManagedCodeGen
                     {
                         Console.WriteLine("Formatting jit directory.");
                     }
-                    _srcDirectory = "src/jit";
+                    _srcDirectory = Path.Combine("src", "jit");
                 }
 
                 if (_projects.Count == 0 && _verbose)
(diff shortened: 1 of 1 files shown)


src/jit-format/jit-format.cs
```diff
@@ -164,7 +164,7 @@ private void validate()
                     {
                         Console.WriteLine("Formatting jit directory.");
                     }
-                    _srcDirectory = "jit";
+                    _srcDirectory = "src/jit";
```
kunalspathak, line 167: `Path.Combine("src", "jit");`

kunalspathak (changes requested):
Minor suggestion.

BruceForstall (approved):
LGTM after fixing Kunal's suggestion.

You can test this locally running jit-format, but the real test requires you to merge it, and then the next dotnet/runtime PR will automatically pick up the change (it clones the jitutils repo)

kunalspathak:
To test, you could locally run `jit-format`:

```
c:\jitutils>bootstrap.cmd
<<builds everything>>

c:\jitutils>bin\jit-format -a x64 -b Checked -c D:\git\dotnet-runtime\src\coreclr --verbose --fix --projects dll
```

src/jit-format/jit-format.cs
trylek: Good point, thanks, fixed in 2nd commit.

kunalspathak (approved):
LGTM. Thanks!

trylek:
@kunalspathak, @BruceForstall - turns out the jit-format tool was actually missing the definition of the command-line argument to let us specify the source path. I have added it in the latest iteration; could you please take another look when you have a chance?

kunalspathak:
> @kunalspathak, @BruceForstall - turns out the jit-format tool was actually missing the definition of the command-line argument to let us specify the source path. I have added it in the latest iteration; could you please take another look when you have a chance?

I don't think it is needed. We set it correctly to `"jit"` folder inside `validate()` method if it is `null` (which will be the case always). Unless I misunderstood your proposal, the changes you had before your latest commit should serve the purpose. When we remove `"src"` directory, we will just change from "Path.Combine("src", "jit")` back to "jit".

trylek:
Hmm, my original thinking was that with the extra parameter the December rollout would be doable by just making a runtime change; without it, the rollout will technically break the build (the formatting leg) and I'll need to "fix it" by making this counterpart change to the jitutils repo. Based on your suggestion I'm just closing this change because there's no point for making it out of sync with the rollout itself - I don't think the two lines of "cleanup" regarding the "src" folder level merit a separate PR, this can be easily fixed as part of the "real" change in lock-step with the rollout.

PR #305, opened 2020-11-26
Fix regex that parses git diff

kunalspathak:
We occasionally see warnings with `jit-analyze`:

```cmd
Couldn't parse output '182160   0       E:/lsra/binaries/EhWriteThru/bugfix/{dasmset_5/diff => dasmset_2/base}/System.Web.HttpUtility.log'.
Couldn't parse output '182160   0       E:/lsra/binaries/EhWriteThru/bugfix/{dasmset_5/diff => dasmset_2/base}/System.Web.log'.
```

They come depending on where `git diff` command was ran. I could see that the output of `git diff` was different depending on where `git diff` is ran, I am not sure why. Here is an example:

```cmd
D:\git\runtime>git diff --no-index --diff-filter=M --exit-code --numstat -- E:\lsra\binaries\EhWriteThru\bugfix\dasmset_5\tempd E:\lsra\binaries\EhWriteThru\bugfix\dasmset_2\tempb
34      51      "E:\\lsra\\binaries\\EhWriteThru\\bugfix\\dasmset_5\\tempd/CommandLine.dasm" => "E:\\lsra\\binaries\\EhWriteThru\\bugfix\\dasmset_2\\tempb/CommandLine.dasm"
182160  0       "E:\\lsra\\binaries\\EhWriteThru\\bugfix\\dasmset_5\\tempd/CommandLine.log" => "E:\\lsra\\binaries\\EhWriteThru\\bugfix\\dasmset_2\\tempb/CommandLine.log"

D:\git\runtime>cd D:\git\jitutils\src\jit-analyze\bin\Debug\net5.0

D:\git\jitutils\src\jit-analyze\bin\Debug\net5.0>git diff --no-index --diff-filter=M --exit-code --numstat -- E:\lsra\binaries\EhWriteThru\bugfix\dasmset_5\tempd E:\lsra\binaries\EhWriteThru\bugfix\dasmset_2\tempb
34      51      E:/lsra/binaries/EhWriteThru/bugfix/{dasmset_5/tempd => dasmset_2/tempb}/CommandLine.dasm
182160  0       E:/lsra/binaries/EhWriteThru/bugfix/{dasmset_5/tempd => dasmset_2/tempb}/CommandLine.log
```

In the second example, the file paths were condensed to use `=>` which fails our parsing logic. So I decided to pass workingDirectory for `git diff` command to be that of `basePath` and that prints complete path in the `git diff` output.

Edit: I have also updated the regex to fix the parsing error, so we should not see them anymore. So, with that we don't need to pass the `basePath` as working directory.

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -994,8 +994,8 @@ namespace ManagedCodeGen
                     return fileToTextDiffCount;
                 }
 
-                string fullBasePath = Path.GetFullPath(basePath);
-                string fullDiffPath = Path.GetFullPath(diffPath);
+                string fullBasePath = Path.GetFullPath(basePath).TrimEnd(Path.DirectorySeparatorChar);
+                string fullDiffPath = Path.GetFullPath(diffPath).TrimEnd(Path.DirectorySeparatorChar);
 
                 for (int i = 0; i < rawLines.Length; i += 3)
                 {
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -972,7 +972,7 @@ namespace ManagedCodeGen
             commandArgs.Add(diffPath);
             commandArgs.Add(basePath);
 
-            ProcessResult result = Utility.ExecuteProcess("git", commandArgs, true);
+            ProcessResult result = Utility.ExecuteProcess("git", commandArgs, true, basePath);
             Dictionary<string, int> fileToTextDiffCount = new Dictionary<string, int>();
 
             if (result.ExitCode != 0)
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -1011,6 +1011,11 @@ namespace ManagedCodeGen
                         };
 
                         string[] splitLine = gitMergedOutputRegex.Split(modifiedLine[2]);
+                        if (splitLine.Length != 6)
+                        {
+                            Console.WriteLine($"Couldn't parse output '{line}`.");
+                            continue;
+                        }
 
                         // Split will output:
                         //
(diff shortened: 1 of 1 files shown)


kunalspathak:
@dotnet/jit-contrib

src/jit-analyze/jit-analyze.cs
```diff
@@ -997,7 +997,7 @@ public static void GenerateTSV(IEnumerable<FileDelta> compareList, string path)
                     // This should be split into:
                     //
                     // 32\t2\t/coreclr/bin/asm/asm/base/101301.dasm\t/coreclr/bin/asm/asm/diff/101301.dasm
-                    Regex gitMergedOutputRegex = new Regex(@"\{(\w+)\s=>\s(\w+)\}");
+                    Regex gitMergedOutputRegex = new Regex(@"\{((\w+(?:\/|\\)?)+)\s=>\s((\w+(?:\/|\\)?)+)\}");
```
jakobbotsch, line 1000: `git diff` has the `-z` command line option which will produce machine readable output (i.e. no =>, no escaped characters, things are separated by \0). Maybe it would be better to start using that to avoid more future problems?

kunalspathak: Thank you for the suggestion. I was not aware of `-z` option and it looks good. With that, I can get rid of some of the regexs and have simpler logic. The only downside is while debugging in VS, the output of `git diff` doesn't show properly in the debugger because of trailing `\0`, but I think that should be OK.

image

kunalspathak:
I have also fixed a bug that was not short circuit diff process if "git diff" returned no difference.

BruceForstall (approved):
LGTM, but prefer not adding additional default output.

src/jit-analyze/jit-analyze.cs
```diff
@@ -948,127 +948,98 @@ public static void GenerateTSV(IEnumerable<FileDelta> compareList, string path)
...
-            // only diff files that are present in both base and diff.
             commandArgs.Add("--diff-filter=M");
             commandArgs.Add("--exit-code");
             commandArgs.Add("--numstat");
-            commandArgs.Add(diffPath);
+            commandArgs.Add("-z");
             commandArgs.Add(basePath);
+            commandArgs.Add(diffPath);
 
+            Console.WriteLine($"git {string.Join(" ", commandArgs)}");
```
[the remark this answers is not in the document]
kunalspathak: sure. For now, I will delete the extra default output.

PR #310, opened 2021-01-11
Remove MyGet references, add legacy feed

garath:
Found in response to dotnet/core-eng#11839. dotnet.myget.org was recently shut down, and these references were missed. 

It appears most references weren't used at all (or packages retrieved from other feeds). The dotnet-corefxlab feed was providing system.commandline.0.1.0-e170320-1, which is now found on the myget-legacy feed. 

Tested locally by running build.cmd to completion.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -3,11 +3,8 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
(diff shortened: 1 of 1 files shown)


AndyAyersMS (approved):
More motivation to move off of this command line parser.

cc @dotnet/jit-contrib

PR #311, opened 2021-01-13
Add AllocSize and ExtraAllocBytes metric in jit-analyze

kunalspathak:
In https://github.com/dotnet/runtime/pull/44370 , I added way to output the allocation size in the disasm. Use that to compare the allocation size. I have also added `ExtraAllocBytes` metric that will be used to diff (alloc size - code size) and in ideal cases it should be 0.


kunalspathak:
@dotnet/jit-contrib

BruceForstall (approved):

Issue #125, opened 2018-01-29
jit-analyze: create filtered and optionally markdown-friendly results file


AndyAyersMS:
The current approach requires on run a file diff on rather large text files (eg 80MB). Since jit-analyze knows which methods have size diffs it should be able to create filtered views that just focus on the methods with diffs. These could be prioritized in various orders, eg largest regressions, largest percent regressions, etc.

Furthermore it would be great if this view could also be written out in some markdown friendly fashion. I've tried two different approaches here, with mixed results.

First is to just create patch files. These often don't have enough context, but they get rendered reasonably well by github.

Second is to create tables with side by side views. Alignment can be tricky and you might lose syntax and diff highlighting but I find this easier to parse, at least in some cases.

Creation of either of these manually is laborious, so automating it would be really nice.

PR #312, opened 2021-01-22
Enable jit-analyze to take multiple metrics as input and add an option to output .md file

kunalspathak:
I was tired of gathering the diff data for individual metric by invoking `jit-analyze` multiple times, so I went ahead and added ability to specify multiple metrics in comma-separated form so we can see the diffs of multiple metrics at one instead of invoking `jit-analyze` multiple times. Another motivation was that I wanted to output the diff result in `.md` file so I can just copy the contents from it and paste it on GH. So I went ahead and added an option of `--md` that will output the diff result (of all the metrics specified) in an `.md ` file.

Example of what you end up getting in `.md` can be seen in https://github.com/dotnet/runtime/pull/47307#issuecomment-765046091.

Contributes to https://github.com/dotnet/jitutils/issues/125

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -38,6 +38,18 @@ namespace ManagedCodeGen
 
     public class jitanalyze
     {
+
+        private const string DETAILS_MARKER = "SUMMARY_MARKER";
+        private static string METRIC_SEP = new string('-', 80);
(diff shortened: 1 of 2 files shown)


kunalspathak:
@dotnet/jit-contrib , @AndyAyersMS

AndyAyersMS (approved):
Looks good.

PR #316, opened 2021-02-15
Fix "error: invalid option --metric" error

EgorBo:
`jit-diff diff` executes `jit-analyze` after running the diffs and always fail with `error: invalid option --metric`
I guess it was introduced in https://github.com/dotnet/jitutils/pull/312


AndyAyersMS (approved):
Thanks.

Suspect many folks have shifted over to using SPMI for diffs.

JulieLeeMSFT:
I had this error too. Will use this fix.

kunalspathak:
Thanks for fixing this.

PR #317, opened 2021-02-16
Add Nullable<int> to pmi

EgorBo:
Makes `--pmi` 7% slower.

@AndyAyersMS did you mean to file a PR like this in https://github.com/dotnet/runtime/pull/48161#issuecomment-777627105 or just to test locally?


AndyAyersMS:
This has come up often enough now that I think we should just add it...

AndyAyersMS (approved):

PR #318, opened 2021-03-08
Improve jit-format

BruceForstall:
1. Currently, if you run `jit-format` but aren't already using
ninja, it does a full ninja build of the dotnet/runtime repo.
Change the invocation of `build-runtime.cmd` to use the
`-configureonly -ninja` args instead of `usenmakemakefiles` so
only the cmake configure part of the build is run, which generates
the `compile_commands.json` that the format tool needs. After this
goes in, we can remove the obsolete `usenmakemakefiles` switch
from `build-runtime.cmd`, as it is only there to support `jit-format`.

2. `jit-format` invokes `dotnet --info` to figure out the OS. But
this doesn't work when invoking the executable dotnet (as opposed
to the script wrapper dotnet.cmd/sh) when inside the runtime repo,
as it doesn't report as much information as we want. So, for simplicity,
just invoke the command in some location (like the Windows system
directory, or Unix root directory) that shouldn't be a runtime repo
clone, so shouldn't have this problem.

Note that this fix should also get applied to `jit-diff` (actually,
it should get refactored into util.cs and shared).

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -85,7 +85,26 @@ namespace ManagedCodeGen
                     Console.WriteLine("Running: {0} {1}", "dotnet", String.Join(" ", commandArgs));
                 }
 
-                ProcessResult result = Utility.ExecuteProcess("dotnet", commandArgs, true);
+                // Running "dotnet" (the executable, not the .cmd/.sh wrapper script) when the current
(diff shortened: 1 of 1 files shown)


BruceForstall:
@dotnet/jit-contrib PTAL

kunalspathak (approved):
Thanks for fixing the `ninja` command.

sandreenko (approved):

PR #320, opened 2021-03-10
Explicitly use the oldest* available LTS framework

jashook:
Explicitly use the oldest* available LTS framework as the TargetFramework. This will more easily allow using tools like PMI with both a 3.1 runtime and a 5.0 runtime.

### DIFF
--- a/src/target-framework.props
+++ b/src/target-framework.props
@@ -1,12 +1,9 @@
 <Project>
-
-  <PropertyGroup Condition="!$(NETCoreSdkVersion.StartsWith('5'))">
-    <_UseNetCoreApp3>true</_UseNetCoreApp3>
-  </PropertyGroup>
-
   <PropertyGroup>
-    <TargetFramework Condition="$(_UseNetCoreApp3) == true">netcoreapp3.1</TargetFramework>
-    <TargetFramework Condition="$(_UseNetCoreApp3) != true">net5.0</TargetFramework>   
+    <!-- Target the oldest possible LTS framework version. This will allow   -->
+    <!-- using multiple versions of the runtime with the same managed        -->
+    <!-- executable.                                                         -->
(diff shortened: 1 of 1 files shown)


jashook:
Note that 2.1 is still in support; however, to avoid changing it in a few months I chose 3.1.

BruceForstall:
Seems reasonable, but what happens if you don't have 3.1 LTS installed (but do have 5.0 installed)? I think we need to support that scenario.

jashook:
The SDK will allow building an older TargetFramework, and it will just run with the newer runtime. Both are supported scenarios.

Example:

> docker run -it mcr.microsoft.com/dotnet/sdk:5.0 bash 
> git clone https://github.com/jashook/jitutils -b user_older_target_framework
> ./build.sh -b release

Will give a clean build with only the 5.0 SDK installed. The output managed dlls are all targeting .netcoreapp3.1; however, can be run with a newer runtime (5.0+).

jashook:
The TargetFramework property should only need to be updated when the *oldest* LTS SDK goes out of support. At that point in time theoretically we would not care to support those runtimes anyways.

jashook:
Gentle ping @BruceForstall

jashook:
Let me know how you would like to go forward, to use the 5.0 sdk/runtime without explictly requiring someone to download the 3.1 runtime would require some sort of parsing of dotnet --list-runtime and passing an env variable to be picked up by msbuild.

kunalspathak:
I agree with what @jashook is proposing.

jashook:
The problem is the following:

1. We have the 5.0 sdk installed
2. Using a 3.1 build of coreclr we are unable to run because pmi.dll will attempt to load System.Runtime 5.0.0.

BruceForstall (approved):
Let's go with it.

PR #322, opened 2021-03-16
Support the nmakeobj path and the standard intermediate output path for compile_commands.json on Windows.

jkoritzinsky:
Required for https://github.com/dotnet/runtime/pull/49715

cc: @dotnet/jit-contrib for review.

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -231,6 +231,7 @@ namespace ManagedCodeGen
                     // If the user didn't specify a compile_commands.json, we need to see if one exists, and if not, create it.
                     if (!_untidy && _compileCommands == null)
                     {
+                        // Check both the nmakeobj and obj paths for back-compat with the old Ninja/NMake output dir.
                         string[] nmakeObjCompileCommandsPathSegments = { _rootPath, "..", "..", "artifacts", "nmakeobj", _os + "." + _arch + "." + _build, "compile_commands.json" };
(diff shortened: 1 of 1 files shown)


BruceForstall (approved):
Is `nmakeObjCompileCommands` still here just to support people who haven't updated to the newest (PR'ed) dotnet/runtime? Or do we check both paths for some other reason, like cmake might put things in a different place in some scenarios? it might be helpful to add the answer as comments in the code.

jkoritzinsky:
Yes, it's just here for backcompat. I'll add a comment in the code.

src/jit-format/jit-format.cs
```diff
@@ -231,11 +231,21 @@ private void validate()
                     // If the user didn't specify a compile_commands.json, we need to see if one exists, and if not, create it.
                     if (!_untidy && _compileCommands == null)
                     {
-                        string[] compileCommandsPath = { _rootPath, "..", "..", "artifacts", "nmakeobj", _os + "." + _arch + "." + _build, "compile_commands.json" };
-                        _compileCommands = Path.Combine(compileCommandsPath);
+                        string[] nmakeObjCompileCommandsPathSegments = { _rootPath, "..", "..", "artifacts", "nmakeobj", _os + "." + _arch + "." + _build, "compile_commands.json" };
```
jkoritzinsky, line 234:
```suggestion
                        // Check both the nmakeobj and obj paths for back-compat with the old Ninja/NMake output dir.
                        string[] nmakeObjCompileCommandsPathSegments = { _rootPath, "..", "..", "artifacts", "nmakeobj", _os + "." + _arch + "." + _build, "compile_commands.json" };
```

PR #327, opened 2021-04-28
Add download retry on failure for clang tools

BruceForstall:
On download failure, add 3 retries, with 15 second waits between them.

### DIFF
--- a/bootstrap.cmd
+++ b/bootstrap.cmd
@@ -16,7 +16,7 @@ where /q git.exe
 if %errorlevel% NEQ 0 echo Can't find git.exe! Please add to PATH&set __ExitCode=1&goto :script_exit
 
 set __root=%~dp0
-setlocal
+setlocal ENABLEEXTENSIONS ENABLEDELAYEDEXPANSION
 
 REM Are we already in the dotnet/jitutils repo? Or do we need to clone it? We look for build.cmd
 REM in the current directory (which is the directory this script was invoked from).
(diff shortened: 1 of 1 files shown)


BruceForstall:
(Hopefully) fixes https://github.com/dotnet/runtime/issues/51943

I can't force the downloads to fail, but I did test it with artificial failures.

@dotnet/jit-contrib PTAL
cc @ViktorHofer @trylek @jkoritzinsky @hoyosjs

bootstrap.cmd
```diff
@@ -145,15 +145,39 @@ echo Tools can be found at http://llvm.org/releases/download.html#3.8.0
...
 
-call powershell Invoke-WebRequest -Uri "https://clrjit.blob.core.windows.net/clang-tools/windows/clang-format.exe" -OutFile bin\clang-format.exe
-if errorlevel 1 echo ERROR: failed to download clang-format.&set __ExitCode=1&goto :eof
+call :download_url clang-format "https://clrjit.blob.core.windows.net/clang-tools/windows/clang-format.exe" bin\clang-format.exe
+if %__ExitCode% NEQ 0 goto :eof
 
-call powershell Invoke-WebRequest -Uri "https://clrjit.blob.core.windows.net/clang-tools/windows/clang-tidy.exe" -OutFile bin\clang-tidy.exe
-if errorlevel 1 echo ERROR: failed to download clang-tidy.&set __ExitCode=1&goto :eof
+call :download_url clang-tidy "https://clrjit.blob.core.windows.net/clang-tools/windows/clang-tidy.exe" bin\clang-tidy.exe
+if %__ExitCode% NEQ 0 goto :eof
```
hoyosjs, line 152: This doesn't change previous behavior, so I'm not saying this has to fix it. But it's not clear to me who uses `__ExitCode` for this subprocess and how that will bubble up to the caller.

hoyosjs (approved):

bootstrap.cmd
BruceForstall: This is the `:build_jitutils` path, and the caller uses `__ExitCode` to set the exit code by calling `:script_exit`

PR #328, opened 2021-05-20
Restrict console output for methods and types with QUIET passed

jashook:
In addition, if JitDisasm is set skip methods which do not have the
method name. This avoids compiling all methods when there is only
one that is intended to be dumped.

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -440,18 +440,70 @@ abstract class PrepareBase : CounterBase
 
         bool useEnv = methodToDisasm != null;
         ReadOnlySpan<char> methodName = null;
+        ReadOnlySpan<char> className = null;
+
+        bool namesMatch = false;
 
         if (useEnv)
         {
-            int methodNameStart = methodToDisasm.IndexOf(':') + 1;
+            bool hasClassName = methodToDisasm.IndexOf(':') != -1;
+            bool containsArgs = methodToDisasm.IndexOf('(') != -1;
+            bool endsWithWildcard = methodToDisasm.IndexOf('*') != -1;
+
+            // COMPlus_JitDisasm=* passed, this is the same as not having
+            // the environment set.
+            if (endsWithWildcard && methodToDisasm.Length == 1)
+            {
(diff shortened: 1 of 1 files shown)


jashook:
@BruceForstall please route as necessary

BruceForstall:
+ @AndyAyersMS

BruceForstall:
The JitDisasm changes worry me, as they're a bit of a hack: they don't parse using the same syntax as the JIT. E.g., they don't handle `*`. And it doesn't look like the two cases parse the same way.

The verbose changes to DriveAll seem reasonable, but I haven't been able to successfully run DriveAll locally to see what the current behavior is, so I need to investigate that first.

jashook:
This is correct, anything with the wildcard will currently fail. Adding logic for that, would be a bit annoying but seems required.

AndyAyersMS:
I've wanted something like this for a while, though generally I'd prefer to specify the pattern via command line args instead of env vars (which I realize is awkward because pmi's command line parsing is lame).

So maybe (in verbose) mode indicate that the results are filtered?

jashook:
I am fine with changing this to be passed in via command line arguments. Seems to somewhat diverge from current expectations; however, seems like the correct direction.

As for the filtered indications, sounds like printing method ------ is skipped.

>I've wanted something like this for a while

I found pretty quickly it is necessary for large assemblies. Between the disk usage and just the time to dasm a large assembly makes this change a requirement for viewing dasm for a particular method in most assemblies.

jashook:
Thank you for the initial comments, I will work on the feedback. This is no where near my highest priority, so my guess is I will touch back on this in a week or two.

BruceForstall:
fwiw, another workflow could be: superpmi collect a PMI run, then do superpmi replay with asm for a particular function or functions. This would be a one-time hit for the collection, but everything subsequent is really fast. I can give more details if interested.

jashook:
I think the larger issue here with relying on SuperPMI is the fact that the assemblies are not snapshots. They change often and recollection is expensive and long.

PR #329, opened 2021-05-21
Add FindRedundantTest() in AnalyzeAsm

kunalspathak:
Added `FindRedundantTest()` method that looks for redundant `test` instruction in the dasm files. I have scoped it to limited jump instructions, but can be extended in the future. Used in https://github.com/dotnet/runtime/issues/53053.

### DIFF
--- a/src/AnalyzeAsm/Program.cs
+++ b/src/AnalyzeAsm/Program.cs
@@ -97,13 +97,25 @@ namespace AnalyzeAsm
             //RedundantMovs2($@"{workingFolder}\redundant-mov-2.asm");
             //RedundantMovs3($@"{workingFolder}\redundant-mov-3.asm");
             //PrintAssembly(armFile, "System.Linq.Expressions.Interpreter.ActionCallInstruction:Run(System.Linq.Expressions.Interpreter.InterpretedFrame):int:this");
-            PrintAssembly();
+            //PrintAssembly();
             // No index
             //var dasmFiles = Directory.GetFiles(armFolderForJit, "*.dasm", SearchOption.TopDirectoryOnly);
             //foreach (var dasmFile in dasmFiles)
             //{
             //    PrintAssembly(dasmFile, "System.Linq.Expressions.Interpreter.ActionCallInstruction:Run(System.Linq.Expressions.Interpreter.InterpretedFrame):int:this");
             //}
+            var dasmFiles = Directory.GetFiles(@"E:\perfinvestigation\52297\binaries\dasm\dasmset_1\base", "*.dasm", SearchOption.TopDirectoryOnly);
+            StringBuilder result = new StringBuilder();
+            int totalMethods = 0, totalTestGroups = 0, foundMethods = 0;
+            foreach (var dasmFile in dasmFiles)
+            {
+                Console.WriteLine($"Processing {dasmFile}");
+                result.Append(FindRedundantTest(dasmFile, ref totalMethods, ref foundMethods, ref totalTestGroups));
+            }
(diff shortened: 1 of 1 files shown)


kunalspathak:
@dotnet/jit-contrib

src/AnalyzeAsm/Program.cs
```diff
@@ -1912,13 +1924,132 @@ static void CSEForAdrpAddAddresses(string inputFile, string outputFile)
...
+            //var aluRegex = new Regex(@"(add|and|dec|inc|lsl|neg|or|sal|sar|shl|shr|sbb|sub|xor)      (\w+),?");
+            //List<string> ops = new List<string>()
+            //{
+            //    " je ", " jne ", " jz ", "jnz "
+            //};
+
+            var aluRegex = new Regex(@"(add|btr|btc|bts|imul|mul|rol|ror|sal|sar|shl|shr|sbb|sub|xor)      (\w+),?");
+            List<string> ops = new List<string>()
+            {
+                " jb ", " jnb ", " jae ", "jnae "
```
AndyAyersMS, line 1950: Why aren't `je`  et al. in this list?

AndyAyersMS: Or did you do two different passes...?

src/AnalyzeAsm/Program.cs
```diff
@@ -1912,13 +1924,132 @@ static void CSEForAdrpAddAddresses(string inputFile, string outputFile)
             WriteResults(outputFile, resultBuilder.ToString());
         }
 
+        /// <summary>
+        /// Summarizes "ldr wzr
```
AndyAyersMS, line 1928: Stale comment?

AndyAyersMS (commented):
Would also be interesting to screen for cases where we have a non-move writing a register, followed by a test, followed by a jump.

In particular `lea; test; jcc`

src/AnalyzeAsm/Program.cs
kunalspathak: I did it in 2 passes. With `jb`, etc. I didn't see any pattern. The patterns that I attached in dotnet/runtime#53053 are from `je`.

kunalspathak:
> Would also be interesting to screen for cases where we have a non-move writing a register, followed by a test, followed by a jump.
> 
> In particular `lea; test; jcc`

I can add more patterns but I just restricted myself to handful of them for time sake. `lea` doesn't affect any flag, so wondering how can `lea` screening help? We won't be able to optimize `test` anyway in that case, right?

AndyAyersMS:
> how can `lea` screening help?

If we can find places where we're forming `lea` upstream of a jump we might want to reconsider and use a flag-setting instruction instead, as we may be happier overall since we can then omit the `test`.

For instance, if we have `lea; test; jmp` we might be better off with `add; jump`.

kunalspathak:
> For instance, if we have `lea; test; jmp` we might be better off with `add; jump`.

So although `lea` is faster than `add`, you are thinking that elimination of extra `test` will offset the degradation?

AndyAyersMS:
Hard to say. Both `lea; test; jcc` and `add; jcc` will benefit from macro-fusion. But I suspect the latter is faster, provided we're only adding one thing in the `lea`.

kunalspathak:
Just found 19 hits of `lea` like these:

```asm
; Assembly listing for method System.MathF:Round(float):float
[0053]       lea      ecx, [rsi+rsi]
[0054]       test     ecx, ecx
[0055]       jne      SHORT G_M2148_IG05
```

BruceForstall (approved):

PR #330, opened 2021-05-28
PMI fixes

BruceForstall:
1. Handle a special case in DRIVEALL when PMI is invoked using the `pmi.exe` tool,
instead of `dotnet.exe` or `corerun.exe`, where we need to remove the
implicit `pmi.dll` argument before spawning the child process.
2. In DRIVEALL, when preparing to spawn the sub-process, we replace the DRIVEALL
command with PREPALL. This replacement needs to be case-insensitive, or else a
command-line like "pmi.exe DriveAll" will fail.
3. Change PrepOne to behave the same as PrepAll w.r.t. the `-QUIET` option.
4. Update the help message with `-Tailcalls` and `-Inlines` options, and note that
for PrepOne, the method index is optional (defaults to zero).
5. Fixed (the weird) arg parsing to search for the leading dashes to the options
(before, `PrepOne-Tailcalls` was treated as `PrepAll-Tailcalls` since it just searched
for `all`).

### DIFF
--- a/src/pmi/PMIDriver.cs
+++ b/src/pmi/PMIDriver.cs
@@ -175,19 +175,47 @@ namespace PMIDriver
                     p.StartInfo.EnvironmentVariables[pair.Key] = pair.Value;
                 }
 
+                Process thisProcess = Process.GetCurrentProcess();
+                string driverName = thisProcess.MainModule.FileName;
+
                 // Fetch our command line. Split off the arguments.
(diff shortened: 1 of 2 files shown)


BruceForstall:
@dotnet/jit-contrib PTAL

BruceForstall:
ping

AndyAyersMS (approved):

PR #331, opened 2021-06-04
Introduce relative score metric in jit-analyze

kunalspathak:
Often times, metric delta of one huge method hides the improvements of other smaller methods. Introduce `relative delta` that will calculate `(diff - base) / base` for individual method and then sum the result to display in the summary.

Sample output:

```
Summary of Perf Score diffs:
(Lower is better)

Total PerfScoreUnits of base: 173075468521.34006
Total PerfScoreUnits of diff: 172839103599.6798
Total PerfScoreUnits of delta: -236364921.66 (-0.14% of base)
Total relative delta: 0.13
    diff is an improvement.
    relative diff is a regression.
```

Also, it turned out that we were recalculating `deltaMetrics` without caching. I introduced private `_deltaMetrics` that will cache the result instead of recomputing.
Lastly, added some `ToString()` methods to get better debugging visualization.

### DIFF
--- a/bootstrap.cmd
+++ b/bootstrap.cmd
@@ -40,7 +40,7 @@ set /P gitroot=<%__tempfile%
 del %__tempfile%
 set gitroot=%gitroot:/=\%
 if not %gitroot:~-1%==\ set gitroot=%gitroot%\
-if not %__root%==%gitroot% (
+if /I not %__root%==%gitroot% (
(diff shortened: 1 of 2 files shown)


kunalspathak:
@dotnet/jit-contrib

BruceForstall (approved):

PR #335, opened 2021-08-20
Recognize curl as a downloading tool

am11:
Chances of having curl on a fresh Linux / macOS install are actually higher than having wget, but even if there is a tie, using available tool in PATH which does the job right is good. This is the only script related to runtime that force the user to install wget.

.. was about to run `brew install wget` then thought I should do this first. 😄

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -42,9 +42,17 @@ function get_host_os {
     esac
 }
 
+_machineHasCurl=
+
 function validate_url {
-    if wget -S --spider $1  2>&1 | grep -e "HTTP/1.1 200 OK";
-    then
+    if (( _machineHasCurl == 1 )); then
+        status="$(curl -sLIo /dev/null "$1" -w '%{http_code}\n')"
+    else
+        response=($(wget -S --spider "$1" 2>&1 | grep "HTTP/"))
+        status="${response[1]}"
+    fi
+
+    if (( status >= 200 || status < 400 )); then
         return 0;
     else
         return 1;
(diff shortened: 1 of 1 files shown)


am11:
cc @BruceForstall, PTAL. `dotnet-install.sh` and other scripts in runtime and arcade provide curl or wget choice. This patch aligns bootstrap.sh with the similar fallback mechanism.

BruceForstall:
This LGTM. Can some @dotnet/jit-contrib Unix guru review?

bootstrap.sh
```diff
@@ -42,9 +42,17 @@ function get_host_os {
...
-    if wget -S --spider $1  2>&1 | grep -e "HTTP/1.1 200 OK";
-    then
+    if (( _machineHasCurl == 1 )); then
+        status="$(curl -sLIo /dev/null "$1" -w '%{http_code}\n')"
+    else
+        response=($(wget -S --spider "$1" 2>&1 | grep "HTTP/"))
+        status="${response[1]}"
+    fi
+
+    if (( status >= 200 || status < 400 )); then
```
echesakovMSFT, line 55: What statuses other than `200` and `301` can indicate that the url is valid?

bootstrap.sh
```diff
@@ -72,24 +79,32 @@ function download_tools {
...
 
     clangTidyUrl=https://clrjit.blob.core.windows.net/clang-tools/${info}/clang-tidy
 
-    if validate_url ${clangTidyUrl} > /dev/null; then
+    if validate_url "$clangTidyUrl" > /dev/null; then
         echo "Downloading clang-tidy to bin directory"
         # download appropriate version of clang-tidy
-        wget --tries 4 --progress=dot:giga ${clangTidyUrl} -O bin/clang-tidy
+        if (( _machineHasCurl == 1 )); then
+            curl --retry 4 --progress-bar --fail "$clangTidyUrl" -o bin/clang-tidy
```
am11, line 101:
```suggestion
            curl --retry 4 --progress-bar --location --fail "$clangTidyUrl" -o bin/clang-tidy
```

BruceForstall:
@echesakovMSFT Any more feedback here?

echesakovMSFT (approved):
LGTM

PR #336, opened 2021-08-23
use retry for downloading clang tools

kunalspathak:
Add retry logic during downloading clang tools in unix. Turns out we only added retry logic for windows in https://github.com/dotnet/jitutils/pull/327/files and not for linux.


kunalspathak:
@dotnet/jit-contrib

PR #339, opened 2021-10-29
Use non-deprecated Azure Pipelines Ubuntu/Windows pool

akoeplinger:
ubuntu-16.04 was removed recently (https://github.com/actions/virtual-environments/issues/3287), bumping to 20.04

win2016 will be removed in March: https://github.com/actions/virtual-environments/issues/4312, bumping to windows-2019

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,13 +9,13 @@ jobs:
   parameters:
     agentOs: Windows_NT
     pool:
-      vmImage: vs2017-win2016
+      vmImage: windows-2019
 
 - template: /eng/build.yml
   parameters:
     agentOs: Linux
     pool:
-      vmImage: ubuntu-16.04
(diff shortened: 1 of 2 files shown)


akoeplinger:
@echesakovMSFT it looks like the Linux build fails because there's a new RID and this doesn't work with the logic in https://github.com/dotnet/jitutils/blob/89c6aba529ca53e60229ff5eb57b620e98c031d6/bootstrap.sh#L73-L80

We could either upload a ubuntu 20.04 version to the blob store or download the earlier version, what do you think?

echesakovMSFT:
I think for now we can fix the script by downloading an earlier version. 
Actually, I didn't know that the clang-tools blobs use RIDs...

akoeplinger:
@echesakovMSFT done

echesakovMSFT (approved):
Thanks @akoeplinger

PR #342, opened 2021-11-05
Catch exceptions if spawning a process fails

BruceForstall:
For example, if running `jit-analyze` but `git` is not on the path,
print a more useful error than an exception stack trace.

### DIFF
--- a/src/util/util.cs
+++ b/src/util/util.cs
@@ -226,7 +226,22 @@ namespace ManagedCodeGen
             }
 
             // Finally, start the process.
-            process.Start();
+            try
+            {
+                process.Start();
+            }
+            catch (System.Exception e)
+            {
+                // Maybe the program we're spawning wasn't found (ERROR_FILE_NOT_FOUND == 2).
+                Console.Error.WriteLine($"Error: failed to start '{name} {startInfo.Arguments}': {e.Message}");
+
+                return new ProcessResult()
+                {
+                    ExitCode = -1,
+                    StdOut = string.Empty,
+                    StdErr = string.Empty
+                };
+            }
 
             if (capture)
             {


BruceForstall:
@dotnet/jit-contrib PTAL

src/util/util.cs
```diff
@@ -226,37 +226,52 @@ public static ProcessResult ExecuteProcess(string name, IEnumerable<string> comm
...
+                    // Set up async reads for stderr and stdout
+                    process.BeginErrorReadLine();
+                    process.BeginOutputReadLine();
                 }
-                if (!string.IsNullOrEmpty(stderr) && (stdout != stderr))
+                else
                 {
-                    Console.Write(stderr);
+                    string stdout = process.StandardOutput.ReadToEnd();
+                    string stderr = process.StandardError.ReadToEnd();
```
jakobbotsch, line 242: I know this is preexisting code, but I wonder if there isn't potential for deadlock here if the child process writes more that can fit in its buffer into stderr.

src/util/util.cs
```diff
@@ -226,37 +226,52 @@ public static ProcessResult ExecuteProcess(string name, IEnumerable<string> comm
...
 
-            return new ProcessResult()
+                return new ProcessResult()
+                {
+                    ExitCode = process.ExitCode,
+                    StdOut = capture ? _outputDataStringBuilder.ToString() : string.Empty,
+                    StdErr = capture ? _errorDataStringBuilder.ToString() : string.Empty
+                };
+            }
+            catch (System.Exception e)
```
jakobbotsch, line 263: I think it would be better to do this more locally around `process.Start()` above so that we don't end up printing confusing errors if any of the other function calls above throw an exception.

src/util/util.cs
BruceForstall: Based on the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standardoutput?redirectedfrom=MSDN&view=net-5.0#System_Diagnostics_Process_StandardOutput), the answer appears to be "yes": it even gives this exact example.

I opened https://github.com/dotnet/jitutils/issues/344 to track.

BruceForstall:
@jakobbotsch I updated the change to only surround process Start with try/catch.

jakobbotsch (approved):

PR #345, opened 2021-11-15
Build coredistools for macOS arm64

echesakovMSFT:
I am sure that this is going to fail building for macOS arm64 and thus requiring us to update and use a newer version of LLVM.
Let's test anyway...


echesakovMSFT:
Confirmed that this requires more work than simply updating the target triples in the script.
I will continue working on this in background.

cc @JulieLeeMSFT @BruceForstall

PR #347, opened 2021-11-16
Add option of retainOnlyTopFiles to jit-analyze

kunalspathak:
`RetainBigDiffFilesOnly` will make `jit-analyze` to delete the `.dasm` files and retain just the top `count` files that contains largest diffs. This will help us in reducing the `.dasm` files size when we upload them as part of CI. By default, this option is off and hence we will never delete those files.

Thoughts?

### DIFF
--- a/src/jit-analyze/jit-analyze.cs
+++ b/src/jit-analyze/jit-analyze.cs
@@ -113,7 +113,7 @@ namespace ManagedCodeGen
                     syntax.DefineOption("skiptextdiff", ref _skipTextDiff,
                         "Skip analysis that checks for files that have textual diffs but no metric diffs.");
                     syntax.DefineOption("retainOnlyTopFiles ", ref _retainOnlyTopFiles,
-                        "Retain only the .dasm files that has top 'count' improvements/regressions. Delete other files. Useful in CI scenario to reduce the upload size.");
+                        "Retain only the top 'count' improvements/regressions .dasm files. Delete other files. Useful in CI scenario to reduce the upload size.");
                     syntax.DefineOption("override-total-base-metric", ref _overrideTotalBaseMetric, ParseDouble,
                         "Override the total base metric shown in the output with this value. Useful when only changed .dasm files are present and these values are known.");
                     syntax.DefineOption("override-total-diff-metric", ref _overrideTotalDiffMetric, ParseDouble,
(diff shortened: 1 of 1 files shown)


kunalspathak:
@BruceForstall - any thoughts?

EgorBo (approved):

BruceForstall (changes requested):
The idea seems good. Some comments/suggestions:

1. We already have a list of the base/diff files (`baseList` / `diffList` in `Main`) in a FileInfo list, that includes the full path. Those paths are copied into the `FileDelta` structure. So it seems like we shouldn't do a directory search for files to delete, we should use the file paths we already have.
2. What if you initialize a member of FileInfo (or FileDelta), `retain = false`, only used if the `retain` option is used, set it to `true` in the `DisplayFileMetric` code, then simply iterate over the lists towards the end of `Main` deleting files where this is `false`?
3. Nit: name arg `-retainOnlyTopFiles` instead? Why? "Diff" implies it wouldn't delete "Base" files (which it would/should). "Top" implies the highest metric value, which might not be the "biggest". Maybe even more general: "-retainReportedFiles", meaning to retain any files on which we report a regression/improvement. We might choose to report both the largest, weighted, code size diffs. But we might also want to report the smallest, which are often the easiest to analyze. In fact, for the purpose of saving things in the CI, it would be quite useful to include both those sets.

kunalspathak:
Done.

kunalspathak:
Related PR where this option will be consumed - https://github.com/dotnet/runtime/pull/61700

kunalspathak:
> But we might also want to report the smallest, which are often the easiest to analyze. In fact, for the purpose of saving things in the CI, it would be quite useful to include both those sets.

I will keep that as a future improvement. Let us monitor and see how big the size of top (largest) subset is and if there is a need to see at smaller subset. I usually just pick the smallest of the top files that are reported.

src/jit-analyze/jit-analyze.cs
```diff
@@ -111,6 +112,8 @@ public Config(string[] args)
                         "Only consider assembly files whose names match the filter");
                     syntax.DefineOption("skiptextdiff", ref _skipTextDiff,
                         "Skip analysis that checks for files that have textual diffs but no metric diffs.");
+                    syntax.DefineOption("retainOnlyTopFiles ", ref _retainOnlyTopFiles,
+                        "Retain only the .dasm files that has top 'count' improvements/regressions. Delete other files. Useful in CI scenario to reduce the upload size.");
```
BruceForstall, line 116: grammar nit:
```suggestion
                        "Retain only the top 'count' improvements/regressions .dasm files. Delete other files. Useful in CI scenario to reduce the upload size.");
```

BruceForstall (approved):

Issue #349, opened 2022-03-01
Publish a new version of CoreDisTools NuGet packages


echesakov:
I realized that I haven't published a new version of CoreDisTools NuGet packages after updating the LLVM to 13.0 https://github.com/dotnet/jitutils/commit/6fedaca9622e8db51e3f92ff533a8fcb1aceb8e3

In addition to that, we would need to update clrjit2 blob with the newer versions of the libraries (if we want to use the new version with superpmi.py). IIRC, only macos-arm64 `libcoredistools.dylib` was added to the blob storage. Is this correct, @kunalspathak?

kunalspathak:
I see it is uploaded for osx-64 and osx-arm64

echesakovMSFT:
Un-assigning myself
cc @BruceForstall

BruceForstall:
I published the new coredistools from build https://dev.azure.com/dnceng/public/_build/results?buildId=1665935&view=artifacts&pathAsName=false&type=publishedArtifacts to the clrjit blob container (within https://clrjit2.blob.core.windows.net/superpmi/libcoredistools/; coredistools is downloaded from there by superpmi.py if it doesn't find an existing coredistools in Core_Root).

Note that we don't build windows-arm coredistools any more, so I left the old one there.

@dotnet/jit-contrib Let me know if you notice any problems.

PR #351, opened 2022-03-15
Update LLVM version to 13.0.1

echesakovMSFT:
* Use LLVM 13.0.1
* Use Visual Studio 2022 for building on Windows
* Use Ubuntu 18.04 containers for building on Linux

Work towards https://github.com/dotnet/jitutils/issues/349

### DIFF
--- a/build-coredistools.cmd
+++ b/build-coredistools.cmd
@@ -53,7 +53,7 @@ if not exist "%BinariesDirectory%" (
 pushd "%BinariesDirectory%"
 
 cmake.exe ^
-    -G "Visual Studio 16 2019" ^
+    -G "Visual Studio 17 2022" ^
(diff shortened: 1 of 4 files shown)


echesakovMSFT:
@BruceForstall PTAL

BruceForstall:
Why not Ubuntu 20.04?

BruceForstall (approved):

echesakovMSFT:
> Why not Ubuntu 20.04?

Isn't 18.04 still supported? We usually build Linux binaries on the oldest supported versions of the OS.

BruceForstall:
> Isn't 18.04 still supported?

Yes, it is. 

> We usually build Linux binaries on the oldest supported versions of the OS.

That was my guess.

PR #354, opened 2022-04-22
Add jit instruction count pintool

jakobbotsch:


jakobbotsch:
cc @dotnet/jit-contrib

src/jit-pintool/README.md
```diff
@@ -0,0 +1,14 @@
...
+[PIN](https://www.intel.com/content/www/us/en/developer/articles/tool/pin-a-dynamic-binary-instrumentation-tool.html)
+to measure throughput of the JIT. The pintool counts the number of instructions
+executed inside the JIT only. Furthermore it has some special support to
+integrate with SuperPMI's metric collection to allow support for diffing
+throughput.
+
+## Building
+The easiest way to build it is to follow PIN's manual and adding the pintool
+here as another example. See the "Building the Example Tools" and "Building Your
+Own Tool" sections in the manual.
```
kunalspathak, line 13: Can you add a link to the manual?

AndyAyersMS (approved):
I wonder if on 32 bit targets it would be possible/faster to use doubles for the counting?

src/jit-pintool/README.md
jakobbotsch: The link is above.

src/jit-pintool/clrjit_inscount.cpp
```diff
@@ -0,0 +1,249 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
kunalspathak, line 1: Is this the appropriate license to include given that it is based on Intel's source code?

jakobbotsch: I'm not gonna deny that I have used some snippets from their examples (mostly function declarations), but most of this pintool is written from scratch. Also both are MIT licensed.

src/jit-pintool/README.md
kunalspathak: For easier discovery, I would do something like:

```suggestion
 ["Building the Example Tools"](https://software.intel.com/sites/landingpage/pintool/docs/98547/Pin/html/index.html#BuildingExamples) and ["Building Your Own Tool"](https://software.intel.com/sites/landingpage/pintool/docs/98547/Pin/html/index.html#BUILDINGTOOLS) sections in the manual.
```

kunalspathak (approved):

jakobbotsch:
> I wonder if on 32 bit targets it would be possible/faster to use doubles for the counting?

Hmm, interesting idea. Let me see if I can try that out quickly. Not sure how well PIN handles floating point math (the hot function calls here end up inlined by PIN).

jakobbotsch:
> I wonder if on 32 bit targets it would be possible/faster to use doubles for the counting?

Unfortunately PIN only allows claiming virtual integer registers, so this would require storing it per-thread in memory somewhere, but when I have tested that it is significantly slower than the virtual register support they have.

src/jit-pintool/README.md
jakobbotsch: I would rather not link to a specific version of a manual. Even manuals from releases of under one year ago are already dead.

BruceForstall (approved):

PR #356, opened 2022-09-27
Add --is-diffs-only and --is-subset-of-diffs

jakobbotsch:
It is confusing for jit-analyze to display this summary when invoked through SPMI since SPMI will only create .dasm files for some contexts. Add some options to control the summary written.


jakobbotsch:
cc @dotnet/jit-contrib PTAL @AndyAyersMS 

Need this for https://github.com/dotnet/runtime/pull/76238.

jakobbotsch:
In the long run it would probably be better to have jit-analyze return something machine readable that superpmi.py could consume and format how it wishes, but it's a bit more work than I want to sign up for now.

kunalspathak (approved):

PR #357, opened 2022-10-20
Upgrade TFM to net6.0

am11:
* Update TFM `netcoreapp3.1` -> `net6.0`
* Replace Newtonsoft with STJ.
* Replace `COMPlus_` env vars with `DOTNET_`
* Separate out `MutateTestRootCommand` and update `System.CommandLine` package (as we have in cg2, r2rdump, etc.).
* Update docs and package versions.

cc @jakobbotsch, @BruceForstall


src/target-framework.props
```diff
@@ -3,7 +3,7 @@
     <!-- Target the oldest possible LTS framework version. This will allow   -->
     <!-- using multiple versions of the runtime with the same managed        -->
     <!-- executable.                                                         -->
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
jakobbotsch, line 6: I think the comment above still applies, doesn't it?

jakobbotsch: @BruceForstall do you see any issues here? I recall having this discussion about .NET Core 3.1 before, but I don't recall whether this allowed us to use PMI on full framework or what the exact reasoning was.

am11: Yup, netcoreapp3.1 will reach its EOL in about a month (and I think the final servicing has been shipped).

Besides, I doubt we are updating jitutils package to latest releases anymore in dotnet/coreclr repo (where release/3.x code resides).

BruceForstall: We were targeting 3.1 at the request of @jashook . I think it's reasonable at this point to update to 6.0 (EOL announcement/reminder: https://devblogs.microsoft.com/dotnet/net-core-3-1-will-reach-end-of-support-on-december-13-2022/)

BruceForstall (approved):
Thanks for the contribution!

PR #358, opened 2022-10-25
Fix formatting instructions using the python script to match current runtime repo behavior

davidwrighton:
I noticed that the python script changed its name, and the example command line arguments are out of date.


BruceForstall (approved):
Thanks!

PR #359, opened 2022-10-26
Readd myget-legacy feed

jakobbotsch:
We are still depending on an ancient version of System.Commandline, 0.1.0-e170320-1, which we get from this feed. Readd it to unblock things, otherwise all our SPMI jobs fail in dotnet/runtime when they try to build jitutils.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -3,7 +3,6 @@
   <packageSources>
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
-
     <add key="dotnet7" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7/nuget/v3/index.json" />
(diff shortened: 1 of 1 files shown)


jakobbotsch:
cc @BruceForstall 

@am11 feel free to take it upon yourself to upgrade the problematic projects to the more recent version of System.CommandLine 😄

BruceForstall (approved):
Does the order here matter? It was previously listed first.

jakobbotsch:
> Does the order here matter? It was previously listed first.

It doesn't seem to, I think because of the exact version we specify that is only found in this feed.

> @am11 feel free to take it upon yourself to upgrade the problematic projects to the more recent version of System.CommandLine 😄

@BruceForstall pointed out that this is #193 and that the System.CommandLine we are relying on now might be quite different than what is called System.CommandLine today.

am11:
Thanks! I think I had older versions cached, clearing the cache revealed the error. 😁

I will try to migrate tools to System.CommandLine v2. 🙂

PR #360, opened 2022-10-26
Revert recent .NET upgrade

jakobbotsch:
jitformat pipeline is still using .NET core 3.1. We need to upgrade it in dotnet/runtime before we upgrade jitutils.


jakobbotsch:
We can maybe just merge https://github.com/dotnet/runtime/pull/77511 instead.

BruceForstall (approved):

BruceForstall:
> We can maybe just merge [dotnet/runtime#77511](https://github.com/dotnet/runtime/pull/77511) instead.

Might as well wait a few weeks until .NET 7 GA is released?

jakobbotsch:
> Might as well wait a few weeks until .NET 7 GA is released?

I think together with #361 it shouldn't be an issue to take that now.

jakobbotsch:
Hmm, actually I think #361 will not help, but I have changed https://github.com/dotnet/runtime/pull/77511 to upgrade to .NET 6 instead. I think we will need to stay with that until we upgrade jitutils to target .NET 7.

jakobbotsch:
Should not be necessary anymore.

PR #361, opened 2022-10-26
Allow jitutils SDK versions to roll forward

jakobbotsch:
Hopefully this should avoid some issues in the future related to upgrading .NET SDKs

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,0 +1,6 @@
+{
+  "sdk": {
+    "version": "6.0.400",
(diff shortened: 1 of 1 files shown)


jakobbotsch:
This is how I have Fuzzlyn set up and haven't run into problems consuming that from dotnet/runtime even with .NET 7 SDKs.

jakobbotsch:
Hmm, I'm not actually sure this necessarily helps that much. I'll close this.

PR #362, opened 2022-11-09
Migrate four tools to commandline v2

am11:
Bring them to the same plan as https://github.com/dotnet/runtime/pull/72082 and https://github.com/dotnet/runtime/pull/76467.

Remaining two are jit-diff and jit-format. I have left them (for now ™️),
as it turned out to be quite a feat to migrate jit-diff to this plan. 😁


jakobbotsch:
Thank you! We will probably need to do a little testing with this before we can merge it.

src/jit-analyze/JitAnalyzeRootCommand.cs
```diff
@@ -0,0 +1,128 @@
...
+        public Option<bool> Recursive { get; } =
+            new(new[] { "--recursive", "-r" }, "Search directories recursively");
+        public Option<string> FileExtension { get; } =
+            new("--file-extension", _ => ".dasm", true, "File extension to look for");
+        public Option<int> Count { get; } =
+            new(new[] { "--count", "-c" }, _ => 20, true, "Count of files and methods (at most) to output in the summary. (count) improvements and (count) regressions of each will be included");
+        public Option<bool> Warn { get; } =
+            new(new[] { "--warn", "-w" }, "Generate warning output for files/methods that only exists in one dataset or the other (only in base or only in diff)");
+        public Option<List<string>> Metrics { get; } =
+            new(new[] { "--metrics", "-m" }, _ =>  new List<string> { "CodeSize" }, true, $"Comma-separated metric to use for diff computations. Available metrics: {MetricCollection.ListMetrics()}");
```
jakobbotsch, line 26: Is it still true that this can be specified with comma-separated syntax?

src/jit-analyze/Program.cs
```diff
@@ -55,140 +40,17 @@ public class jitanalyze
...
-            public bool IsSubsetOfDiffs => _isSubsetOfDiffs;
-
-            public string Filter {  get { return _filter; } }
-
-            public List<string> Metrics {  get { return _metrics; } }
-            public bool SkipTextDiff { get { return _skipTextDiff;  } }
-            public bool RetainOnlyTopFiles { get { return _retainOnlyTopFiles; } }
+            _reconcile = !Get(command.NoReconcile);
+            _filter = Get(command.Filter);
+            _overrideTotalBaseMetric = Get(command.OverrideTotalBaseMetric);
```
jakobbotsch, line 49: Does this properly parse with invariant culture?

am11: Good catch! I have added a custom invariant culture parser for `Option<double>`, with the error message which they normally show (e.g. for `int` options).

src/jit-analyze/JitAnalyzeRootCommand.cs
jakobbotsch: Note that superpmi.py in dotnet/runtime is passing this arg so it will need changes too.

It does introduce some weirdness if you try to run superpmi on the servicing branches, but I'm not sure if we really need to worry about it (running superpmi.py on servicing branches is rare, and so is specifying custom metrics to superpmi.py). I would be fine with changing the syntax here, but also cc @dotnet/jit-contrib for a chance to express their concerns.

am11:
Bumped version (based on https://github.com/dotnet/runtime/pull/78577).

jakobbotsch:
> Bumped version (based on https://github.com/dotnet/runtime/pull/78577).

Thanks! I hope to take a look at and do some testing of this PR this week.

jakobbotsch:
The first problem I hit is that System.CommandLine.dll published into the `bin` dir gets overridden by the System.CommandLine.dll, so after `./bootstrap.cmd` the executables output to `bin` that use the new System.CommandLine.dll do not work.

I'll see if there's an easy way to rename one of the DLLs

am11:
Publishing as single-file resolved the conflicting dependencies issue when publishing to the same directory (which is generally not recommended): 95056b2.

jakobbotsch:
> Publishing as single-file resolved the conflicting dependencies issue when publishing to the same directory (which is generally not recommended): https://github.com/dotnet/jitutils/commit/95056b2ab08f977e7ed6e9f0ed836bcc98542cc7.

That should be fine, but I think we need to skip it for PMI that is meant to be loaded into a runtime with a checked JIT.

am11:
I've skipped PMI from PublishSingleFile.

> but I think we need to skip it for PMI that is meant to be loaded into a runtime with a checked JIT.

Just curious, what prevents it from working with singlefilehost? Do we unload/reload the corelib which is not possible with single-file mode or something else?

jakobbotsch:
> Just curious, what prevents it from working with singlefilehost? Do we unload/reload the corelib which is not possible with single-file mode or something else?

The idea of PMI is to execute it with a custom build of the runtime/JIT. It then loads a bunch of assemblies and forcefully JITs lots of functions, which allows producing the disassembly for these functions so that it can be diffed and compared. So we generally don't want to use it with the public runtime that single file would entail.

jakobbotsch:
FWIW, jit-dasm-pmi can be singlefile. Only pmi needs to be framework dependent.

am11:
Thanks for testing it @jakobbotsch. :)

I have updated the condition to only exclude `src/pmi` from `PublishSingleFile`.

src/pmi/pmi.cs
```diff
@@ -174,7 +174,7 @@ public CustomLoadContext(string assemblyPath)
         // Add in current assembly path and framework path
         pmiPath += Path.GetDirectoryName(assemblyPath);
         pmiPath += ";";
-        pmiPath += Path.GetDirectoryName(typeof(object).Assembly.Location);
+        pmiPath += Path.GetDirectoryName(System.AppContext.BaseDirectory);
```
jakobbotsch, line 177: I think this change should be reverted, I think it is going to give the directory of PMI but it should be the directory of the libraries for the runtime under test.

am11:
```suggestion
        pmiPath += Path.GetDirectoryName(typeof(object).Assembly.Location);
```

src/jit-dasm/jit-dasm.cs
```diff
@@ -1,723 +0,0 @@
...
-
-                syntax.DefineOption("r|recursive", ref _recursive, "Scan directories recursively.");
-                syntax.DefineOptionList("p|platform", ref _platformPaths, "Path to platform assemblies");
-                var methodsArg = syntax.DefineOptionList("m|methods", ref _methods,
-                    "List of methods to disasm.");
-                methodsArg.IsHidden = true;
-
-                // Warning!! - Parameters must occur after options to preserve parsing semantics.
-
-                syntax.DefineParameterList("assembly", ref _assemblyList, "The list of assemblies or directories to scan for assemblies.");
```
jakobbotsch, line 84: I don't think this behavior is preserved. From my understanding, the parameter list here means that "unnamed" arguments will go into this parameter, e.g. `jit-dasm --option1 value1 --option2 value2 assembly1 assembly2` will put `assembly1` and `assembly2` into this list. However, on this PR, when I try:
```
jit-diff diff --output C:\dev\temp\diffs\ --tag r2rchange --base C:\dev\dotnet\runtime\artifacts\bin\coreclr\windows.x64.Checked\ --base C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\ --core_root "C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files" --crossgen C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\crossgen2\crossgen2.exe -v
```

(after building a checked build in C:\dev\dotnet\runtime, and copying Core_Root to Core_Root_files) I get

```scala
❯ jit-diff diff --output C:\dev\temp\diffs\ --tag r2rchange --base C:\dev\dotnet\runtime\artifacts\bin\coreclr\windows.x64.Checked\ --base C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\ --core_root "C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files" --crossgen C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\crossgen2\crossgen2.exe -v
No assemblies specified; defaulting to corelib
After setting defaults:
--core_root C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files
--crossgen C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\crossgen2\crossgen2.exe
--output C:\dev\temp\diffs\
--base C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\
Beginning Crossgen CodeSize Diffs for System.Private.CoreLib.dll
Scanning: C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files\System.Private.CoreLib.dll
Started dasm command "--platform C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files --verbose --crossgen C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\crossgen2\crossgen2.exe --jit C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\clrjit.dll --output C:\dev\temp\diffs\r2rchange\base C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files\System.Private.CoreLib.dll"
Dasm command: jit-dasm --platform C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files --verbose --crossgen C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\crossgen2\crossgen2.exe --jit C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\clrjit.dll --output C:\dev\temp\diffs\r2rchange\base C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files\System.Private.CoreLib.dll
Description:
  Managed codegen diff tool (crossgen/AOT)

Usage:
  jit-dasm [options]

Options:
  --altjit <altjit>          If set, the name of the altjit to use (e.g., clrjit_win_arm64_x64.dll)
  -c, --crossgen <crossgen>  The crossgen or crossgen2 compiler exe. []
  -j, --jit <jit>            The full path to the jit library []
  -o, --output <output>      The output path
  -f, --file <file>          Name of file to take list of assemblies from. Both a file and assembly list can be used
  --gcinfo                   Add GC info to the disasm output
  --debuginfo                Add Debug info to the disasm output
  --verbose                  Enable verbose output
  --nodiffable               Generate non-diffable asm (pointer values will be left in output)
  -r, --recursive            Search directories recursively
  -p, --platform <platform>  Path to platform assemblies
  -m, --methods <methods>    List of methods to disasm
  --assembly <assembly>      The list of assemblies or directories to scan for assemblies
  -w, --wait                 Wait for debugger to attach
  -v, --version              Show version information
  -?, -h, --help             Show help and usage information


Unrecognized command or argument 'C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files\System.Private.CoreLib.dll'.

Dasm command "jit-dasm --platform C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files --verbose --crossgen C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\crossgen2\crossgen2.exe --jit C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root\clrjit.dll --output C:\dev\temp\diffs\r2rchange\base C:\dev\dotnet\runtime\artifacts\tests\coreclr\windows.x64.Checked\Tests\Core_Root_files\System.Private.CoreLib.dll" returned with 1 failures
```

jakobbotsch: After fixing the above there is another issue with obtaining a full set of jit disasm with crossgen2, introduced by me recently. https://github.com/dotnet/runtime/pull/80371 should fix it.

am11: Could you push that change to PR branch?

build.sh
```diff
@@ -41,7 +37,10 @@ while getopts "hpfb:" opt; do
         buildType=$OPTARG
         ;;
     p)  
-        publish=true
+        publish=1
+        ;;
+    *)  echo "ERROR: unknown argument %opt"
```
am11, line 42:
```suggestion
    *)  echo "ERROR: unknown argument $opt"
```

am11:
Thanks for the fixes, @jakobbotsch!

If possible, I think it would be nice to move jitutils under https://github.com/dotnet/runtime/tree/main/src/tools and use live build in the CI; diffs legs, for example, can build these tools spending two (or less) additional minutes per run. That will save round-trips and keep things up-to-date with convenience.

jakobbotsch:
> Thanks for the fixes, @jakobbotsch!

No problem, thanks for the help! Sorry for the slowness. I think this is good to go now, I did some last tests and things look ok for me locally now.

> If possible, I think it would be nice to move jitutils under https://github.com/dotnet/runtime/tree/main/src/tools and use live build in the CI; diffs legs, for example, can build these tools spending two (or less) additional minutes per run. That will save round-trips and keep things up-to-date with convenience.
> 
> Edit: it currently clones this repo and use the live build, so moving code to runtime repo will actually save a few seconds. 😅

It has been discussed quite a few times, but so far it has not amounted to something we want to spend time on. There are both good arguments for and against, and if we did it we would probably only move some of the tools since this repo serves as kind of a kitchen sink for a lot of very unpolished JIT tools that we don't want to bloat dotnet/runtime with.

jakobbotsch (approved):

src/jit-analyze/JitAnalyzeRootCommand.cs
jakobbotsch: Do you want to submit the fix for this too? Specifically around here: https://github.com/dotnet/runtime/blob/db717e30839c532cad5b269ac11c5d2c91dad639/src/coreclr/scripts/superpmi.py#L1790-L1791
Looks like the help message for the `-metrics` arg in superpmi.py needs an update too.

PR #363, opened 2022-11-17
Fix jit-diff with crossgen2

jakobbotsch:
* Update references Ngen environment variables
* crossgen2 was invoked with --outputfilepath, but the actual command line arg is --out. This was apparently not an error until recent crossgen2 parser changes.


jakobbotsch:
cc @dotnet/jit-contrib

BrianBohe (approved):

PR #368, opened 2023-04-13
Print the clang tools download url

kunalspathak:
The `clang-format` and `clang-tidy` tool is failing to download although `validate_url` is not complaining about the url. I am guessing the latest dotnet arcade in https://github.com/dotnet/runtime/pull/84698 might have broken the `dotnet --info | grep RID` output because of which we might not be populating the url correctly. Added some logging.

Related: https://github.com/dotnet/runtime/issues/84780


kunalspathak:
Fixed by https://github.com/dotnet/runtime/pull/84799

PR #371, opened 2023-04-18
Build with latest OS build agents and .NET 6 SDK

BruceForstall:


BruceForstall:
@dotnet/jit-contrib PTAL -- first step to get the jitutils CI working again.

I'll fix the coredistools build separately.

tannergooding (approved):
Do any of the projects need to be upgraded to target net6 as well?

BruceForstall:
I think the projects were already updated to target net6.

PR #375, opened 2023-07-02
Update System.CommandLine version

am11:
Other dotnet repos have recently updated command-line-api version following the breaking changes after the API review: https://github.com/dotnet/command-line-api/issues/2146. Lets bring jitutils to the same plan now, as these conversion tweaks are fresh in our muscle memory.

cc @jakobbotsch, @adamsitnik

### DIFF
--- a/src/cijobs/Program.cs
+++ b/src/cijobs/Program.cs
@@ -69,15 +69,13 @@ namespace ManagedCodeGen
             return await CopyAsync(cic);
         }
 
-        private T Get<T>(Option<T> option) => _command.Result.GetValue(option);
+        private T Get<T>(CliOption<T> option) => _command.Result.GetValue(option);
 
-        private static Task<int> Main(string[] args) =>
-            new CommandLineBuilder(new CIJobsRootCommand(args))
-                .UseVersionOption("--version", "-v")
-                .UseHelp()
-                .UseParseErrorReporting()
-                .Build()
-                .InvokeAsync(args);
+        private static int Main(string[] args) =>
+            new CliConfiguration(new CIJobsRootCommand(args).UseVersion())
+            {
+                EnableParseErrorReporting = true
+            }.Invoke(args);
 
         // List jobs and their details from the given project on .NETCI Jenkins instance.
         // List functionality:
--- a/src/cijobs/Program.cs
+++ b/src/cijobs/Program.cs
@@ -71,11 +71,11 @@ namespace ManagedCodeGen
 
         private T Get<T>(CliOption<T> option) => _command.Result.GetValue(option);
 
-        private static int Main(string[] args) =>
+        private static Task<int> Main(string[] args) =>
             new CliConfiguration(new CIJobsRootCommand(args).UseVersion())
             {
                 EnableParseErrorReporting = true
-            }.Invoke(args);
+            }.InvokeAsync(args);
 
         // List jobs and their details from the given project on .NETCI Jenkins instance.
         // List functionality:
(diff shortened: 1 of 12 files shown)


src/cijobs/Program.cs
```diff
@@ -69,15 +69,13 @@ public async Task<int> RunAsync(string name)
...
-                .UseParseErrorReporting()
-                .Build()
-                .InvokeAsync(args);
+        private T Get<T>(CliOption<T> option) => _command.Result.GetValue(option);
+
+        private static int Main(string[] args) =>
+            new CliConfiguration(new CIJobsRootCommand(args).UseVersion())
+            {
+                EnableParseErrorReporting = true
+            }.Invoke(args);
```
adamsitnik, line 78: We don't have a good story for it yet, but in general we want to avoid or even forbid the users to mix sync and async. Since all of the sub commands define async actions here, we should call `InvokeAsync` here as well

```suggestion
        private static Task<int> Main(string[] args) =>
            new CliConfiguration(new CIJobsRootCommand(args).UseVersion())
            {
                EnableParseErrorReporting = true
            }.InvokeAsync(args);
```

adamsitnik (commented):
Overall it looks great, I would change only one thing (PTAL at my comment).

Thank you for your help @am11 !

src/cijobs/Program.cs
am11: Nice tip, I'll try to remember it. Thank you! :)

adamsitnik (approved):
LGTM, thank you @am11 !!

adamsitnik:
@jakobbotsch Is there something more that we should do before merging this PR?

The CI is green and the changes LGTM, but I am simply not familiar with how this repo works and how it's used by other repos.

src/jit-analyze/JitAnalyzeRootCommand.cs
```diff
@@ -9,90 +9,90 @@
...
+        public CliOption<string> MD { get; } =
+            new("--md") { Description = "Dump analysis data to specified file in markdown format" };
+        public CliOption<string> Filter { get; } =
+            new("--filter") { Description = "Only consider assembly files whose names match the filter" };
+        public CliOption<bool> SkipTextDiff { get; } =
+            new("--skip-text-diff") { Description = "Skip analysis that checks for files that have textual diffs but no metric diffs" };
+        public CliOption<bool> RetainOnlyTopFiles { get; } =
+            new("--retain-only-top-files") { Description = "Retain only the top 'count' improvements/regressions .dasm files. Delete other files. Useful in CI scenario to reduce the upload size" };
+        public CliOption<double> OverrideTotalBaseMetric { get; } =
+            new("--override-total-base-metric") { DefaultValueFactory = result =>
```
am11, line 45:
```suggestion
            new("--override-total-base-metric") { CustomParser = result =>
```

src/jit-analyze/JitAnalyzeRootCommand.cs
```diff
@@ -9,90 +9,90 @@
...
 
-                 result.ErrorMessage = $"Cannot parse argument '{optionValue}' for option '--override-total-base-metric' as expected type '{typeof(double).FullName}'.";
+                 result.AddError($"Cannot parse argument '{optionValue}' for option '--override-total-base-metric' as expected type '{typeof(double).FullName}'.");
                  return 0;
-            }, false, "Override the total base metric shown in the output with this value. Useful when only changed .dasm files are present and these values are known");
-        public Option<double> OverrideTotalDiffMetric { get; } =
-            new("--override-total-diff-metric", result =>
+            }, Description = "Override the total base metric shown in the output with this value. Useful when only changed .dasm files are present and these values are known" };
+        public CliOption<double> OverrideTotalDiffMetric { get; } =
+            new("--override-total-diff-metric") { DefaultValueFactory = result =>
```
am11, line 55:
```suggestion
            new("--override-total-diff-metric") { CustomParser = result =>
```

jakobbotsch:
> @jakobbotsch, thanks for the test, it is fixed. Could you try with the last commit? :)

Yep, works now. Thanks!

PR #378, opened 2023-07-25
Add a retry loop in WriteAndFlushNextMethodToPrepMarker

jakobbotsch:
I frequently hit "The process cannot access the file" errors when PMI tries to write the marker file, presumably due to AV software locking this file. This adds a retry loop to try writing the marker a few times before aborting.


jakobbotsch:
cc @dotnet/jit-contrib

BruceForstall (approved):

PR #379, opened 2023-08-17
Add option to enable Linux cross builds for jit-format.

c272:
Currently, when jit-format attempts to build a fresh version of compile-commands.json on Linux, it will always attempt a cross build, regardless of whether this is accurate for the host system. This currently breaks our internal formatting CI, as there is no way to disable this with an option/configuration flag. 

This patch adds an option to enable cross builds when on Linux, rather than defaulting to always building with -cross.

I think this will likely break the current .NET CI though, as viewing the [prior patch](https://github.com/dotnet/jitutils/commit/0eb1143e2e6c85d1ea3a999c1be627473a9f7e50) which introduced this argument indicates this was a quick fix to unblock CI after it moved to using rootfs on Mariner, requiring -cross. If this patch is applied, `--linux-cross` would have to be passed.

### DIFF
--- a/src/jit-format/jit-format.cs
+++ b/src/jit-format/jit-format.cs
@@ -40,6 +40,7 @@ namespace ManagedCodeGen
             private string _srcDirectory = null;
             private bool _untidy = false;
             private bool _noformat = false;
+            private bool _linuxCross = false;
             private bool _fix = false;
             private bool _verbose = false;
             private bool _ignoreErrors = false;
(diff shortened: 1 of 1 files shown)


a74nh:
This bug is currently a little annoying as we (Arm) build everything on the target machines - we do no cross building at all. So, it's currently awkward to run the format checking test.

BruceForstall:
@dotnet/jit-contrib

BruceForstall:
I've written a fix that makes "--cross" a required parameter on platforms that need it (like is done here), and also fixed the CI system:

https://github.com/dotnet/jitutils/pull/380
https://github.com/dotnet/runtime/pull/92751

I'm going to close this in favor of those PRs.

PR #382, opened 2023-10-03
Add instructions retired explorer to jitutils

AndyAyersMS:
Seems like it makes sense to include this here rather than keeping it in my own private repo.

We might consider renaming it, since it only optionally looks for instructions retired; mainly its job is to parse ETW and summarize the profile data and the jit time information.

It also knows about BDN intervals and so is especially useful for summarizing profiles gathered with BDN's `-p ETW` option.


AndyAyersMS:
PTAL @dotnet/jit-contrib

BruceForstall (approved):
I suggest adding a README.md file briefly describing it and possibly giving an example or two of a sample workflow using it.

You might also want to update the jitutils root README.md to mention and link to the new tools you're adding to jitutils.

AndyAyersMS:
> I suggest adding a README.md file briefly describing it and possibly giving an example or two of a sample workflow using it.
> 
> You might also want to update the jitutils root README.md to mention and link to the new tools you're adding to jitutils.

Did those things.

PR #383, opened 2023-10-03
Update for new categories of tiered compilation

AndyAyersMS:
Fix instructions retired explorer to understand new categories of jitted code introduced in .NET 8.

Use proper event field rather than trying to decode flags.

Update trace event to a version with the new OptimizationTier values.


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib 

Fixes some issues I ran into trying to run this on a new machine. For some reason the old way of parsing flags no longer worked, and this new way is cleaner anyways.

Also updates IRE to understand new categories of jitted code it might run across in .NET 8.

BruceForstall (approved):

PR #385, opened 2023-11-01
Add `jit-tp-analyze`

SingleAccretion:
This is a C# port of the `analyze-pin-trace-diff.ps1` script used to produce the summary of detailed (trace) PIN diffs.

It has been noted in the past that such tooling would be beneficial to have in jitutils; this is a start.


SingleAccretion:
@dotnet/jit-contrib

src/jit-tp-analyze/Program.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        // Now create a list of functions which contributed to the difference.
+        long totalAbsInsCountDiff = 0;
+        List<FunctionDiff> diffs = new();
+        foreach (string functionName in allRecordedFunctions)
+        {
+            long diffInsCount = diffTrace.GetValueOrDefault(functionName);
+            long baseInsCount = baseTrace.GetValueOrDefault(functionName);
+            long insCountDiff = diffInsCount - baseInsCount;
+            if (insCountDiff is 0)
```
BruceForstall, line 59: Does this work? What's the difference between this and:
```suggestion
            if (insCountDiff == 0)
```

BruceForstall (approved):
Looks great. Thanks for doing this!

src/jit-tp-analyze/Program.cs
SingleAccretion: It does, this is the same pattern matching as `x is null`. There is no difference for primitives. I've changed it back to `==`.

PR #386, opened 2024-01-10
Enable all AArch64 architecture features

BruceForstall:


BruceForstall:
This allows JitLateDisasm to disassemble more SVE instructions than just those supported by Neoverse-N2. Last I looked, "+all" still had a bug such that it doesn't actually understand/disassemble *all* instructions, but it's much better.

a74nh (approved):
This looks like a good idea to me.

BruceForstall:
Ping

kunalspathak (approved):
LGTM, sorry I forgot about this.

PR #389, opened 2024-01-25
Initial version of a tool for ML on CSEs

AndyAyersMS:
Add a tool that can use ML techniques to explore the JIT's CSE heuristic. Some parts of this are very specific to CSEs, others are general and could be repurposed for use with other heuristics.

This is still work in progress.

Depends on jit changes in dotnet/runtime#96880

### DIFF
--- a/src/jit-rl-cse/MLCSE.cs
+++ b/src/jit-rl-cse/MLCSE.cs
@@ -1,0 +1,2158 @@
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@EgorBo PTAL
cc @dotnet/jit-contrib

BruceForstall (approved):
Understood that it is work-in-progress and non-production. Eventually, if it will be used more than just the initial time, it would be valuable to have more documentation, e.g., examples of use, and presumably would need better parameterization (not hard-coded paths).

PR #390, opened 2024-02-04
Adding a fallback for the platform if it couldn't be detected from dotnet --info

tannergooding:
Running `jit-diff` from the runtime root can fail due to the existence of the `globa.json` and custom SDK that is setup since the `Platform` line may not be printed in that scenario.

This adds some basic fallback logic to ensure we can still work in that scenario.


tannergooding:
CC. @dotnet/jit-contrib

jakobbotsch (approved):

PR #391, opened 2024-02-07
Fix RLCSE for new metrics format

AndyAyersMS:
Parsing was broken when I added the initial `; ` to the metrics line. Fix this.

Also fix MCMC forgetting mode, so you can first run MCMC to explore the range of possible CSEs, then try Policy Gradient to see if it can come close to this.


AndyAyersMS:
@EgorBo PTAL
cc @dotnet/jit-contrib

EgorBo (approved):

PR #395, opened 2024-02-10
RLCSE: fix MCMC and GatherFeatures, overwrite dumps

AndyAyersMS:
The JIT will append dumps to existing files, so using RLCSE to save dumps was creating large files (each sequence's dump is periodically updated to show the impact of the current parameters).

Also I was going to make MCMC and such subcommands, but changed my mind, and forgot to hook the options back up.


AndyAyersMS:
@EgorBo PTAL
cc @dotnet/jit-contrib

EgorBo (approved):

PR #396, opened 2024-02-12
RLCSE: fixes for running on linux

AndyAyersMS:
Also add ability to save dumps of greedy policy behavior.

### DIFF
--- a/src/jit-rl-cse/MLCSE.cs
+++ b/src/jit-rl-cse/MLCSE.cs
@@ -85,15 +85,16 @@ public class MLCSE
         SPMI.spmiCollection = Get(s_commands.SPMICollection) ?? spmiCollection;
         SPMI.checkedCoreRoot = Get(s_commands.CheckedCoreRoot) ?? checkedCoreRoot;
         SPMI.showLaunch = Get(s_commands.ShowSPMIRuns);
+        dumpDir = Get(s_commands.OutputDir) ?? dumpDir;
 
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@EgorBo PTAL
cc @dotnet/jit-contrib

src/jit-rl-cse/MLCSE.cs
```diff
@@ -675,6 +682,43 @@ static void PolicyGradient(IEnumerable<Method> methods)
                                     string dasmRun = SPMI.Run(method.spmiIndex, dasmOptions);
                                     sw.WriteLine($" ---> saved baseline dasm to {baseDasmFile}");
                                 }
+
+                                string greedySequence = "greedy";
+                                string greedyDumpFile = Path.Combine(dumpDir, $"dump-{method.spmiIndex}-{greedySequence}-{r}.d");
+
+                                if (File.Exists(greedyDumpFile))
```
EgorBo, line 689: nit: `File.Delete(file)` doesn't throw when file doesn't exist, so you don't need `Exists` check in such cases 🙂

EgorBo (approved):

src/jit-rl-cse/MLCSE.cs
AndyAyersMS: Thanks. I'll be messing with this code more in the future, I want it to be able to spot when it has problems and automatically collect the right dumps, instead of me having to tell it which ones I want.

PR #397, opened 2024-02-16
RLCSE: streaming SPMI mode

AndyAyersMS:
Add support to use streaming SPMI servers. When enabled, RLCSE will launch one SPMI server process per core. Work dispatched to servers picks an idle server from the pool of servers. This greatly reduces the overhead of doing single-method evaluations.

SPMI streaming mode requires the changes in dotnet/runtime#98440.

Update the main Policy graident work loop and the greedy status loop to use the streaming mode, if enabled. MCMC can leverage this too, but I haven't done that just yet.

Add the ability to track server status.

Add the ability to log the server activity. The intent is that these logs can be replayed by a server instance to diagnose problems. This was useful during bring-up.

Add a running log of the Policy Gradient parameter values, so we can see how they change over time.

Refactor out some of the code in the main Policy Gradient loops.

### DIFF
--- a/src/jit-rl-cse/MLCSE.cs
+++ b/src/jit-rl-cse/MLCSE.cs
@@ -1,9 +1,6 @@
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.CommandLine;
-using System.CommandLine.Parsing;
 using System.Diagnostics;
-using System.IO;
-using System.Linq;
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
@EgorBo PTAL
cc @dotnet/jit-contrib

EgorBo (approved):

PR #401, opened 2024-02-24
MLCSE: fix a few issues

### LINKED ISSUE #400: MLCSE failure parsing baseline stats
We see this metrics line in the coreclr_tests collection:
```
; Total bytes of code 192, prolog size 12, PerfScore 50.50, instruction count 48, allocated bytes for code 192, num cse 6 num cand 1 Standard CSE Heuristic seq 4,2,6,1,5,0,1,0 spmi index 547808 (MethodHash=6d6c4ab2) for method Runtime_62108:LeafMethod6():int:this (FullOpts) 
```
and get confused.

This is an opt repeat method where in the first pass we did 5 CSEs, and in the second pass one CSE.

https://github.com/dotnet/runtime/blob/f32c428c86b4cc41e88e2e5a750c37dfb354e33a/src/tests/JIT/Regression/JitBlue/Runtime_62108/Runtime_62108.csproj#L10

Not really sure how to handle this... I can add code to tolerate it for now but this will lead to confused numbers for that method. If opt repeat becomes common then we'll have to find a better way.

AndyAyersMS:
Screen out methods that were jitted with OptRepeat, their CSE metrics may not make any sense.

For streaming SPMI, increase the wait interval after each wait, as OSX seems to like to put the server processes to sleep right away.

Closes #400.

### DIFF
--- a/src/jit-rl-cse/MLCSE.cs
+++ b/src/jit-rl-cse/MLCSE.cs
@@ -1621,7 +1621,7 @@ public static class CollectionData
         if (!File.Exists(cseIndexFile))
         {
             Console.Write("building CSE index for collection ... ");
-            // Sun SPMI (parallel) across entire collection
+            // Run SPMI (parallel) across entire collection
             // and gather the metrics
             string indexContents = SPMI.Run(null, null);
             Console.WriteLine("done");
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@EgorBo PTAL
cc @dotnet/jit-contrib 

I should probably revise the streaming mode implementation since it can be unfair, but this will shore things up a bit longer.

image

EgorBo (approved):

PR #402, opened 2024-03-01
RLCSE: add some validation, consistently divide by base

AndyAyersMS:
Add some sanity checks that the SPMI collection and the SPMI tool are of the same vintage. It would be nice if MCHs conveyed this information directly but for typical usage we can rely in the guid being there in the file name.

Would be nice to do similar for collection Os/Arch but am not doing that yet.

Attempt to create the output dir if it does not exist.

Consistently divide by the base value when comparing against best. This means good comparison values are always less than 1, and bad one greater than one.


AndyAyersMS:
@EgorBo PTAL
cc @dotnet/jit-contrib @NoahIslam @LogarithmicFrog1

EgorBo (approved):

PR #405, opened 2024-03-16
Limit jit-diff parallelism

MihaZupan:
Currently, jit-diff will start jit-dasm processes for all assemblies at the same time.
This change limits how many processes we start at the same time to avoid overwhelming machines with less memory available.

Main motivation is that I'm currently forced to use `--sequential` with @MihuBot when running on Azure Container Instances since they're quite resource limited.


EgorBo (approved):
LGTM

PR #407, opened 2024-04-01
Update jit-format to work on arm64 Mac

BruceForstall:
The current version of clang-format/clang-tidy that we're using (3.8) doesn't have an arm64 Mac version, but this allows jit-format to work with newer clang-format/clang-tidy that do work on arm64 Mac.


BruceForstall:
@dotnet/jit-contrib PTAL

BruceForstall:
Subsumed by #408

PR #412, opened 2024-09-17
Try new Azure Linux containers

BruceForstall:
- Update to the newest Azure Linux 3 containers used by .NET 10 dotnet/runtime builds
- Update to LLVM 20.1.0

### DIFF
--- a/build-coredistools.cmd
+++ b/build-coredistools.cmd
@@ -2,7 +2,7 @@
 setlocal EnableDelayedExpansion EnableExtensions
 
 set TargetOSArchitecture=%1
-set LLVMTargetsToBuild=AArch64;ARM;X86
+set LLVMTargetsToBuild=AArch64;ARM;X86;LoongArch;RISCV
 
 if /i "%TargetOSArchitecture%" == "win-arm64" (
     set GeneratorPlatform=ARM64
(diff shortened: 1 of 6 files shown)


BruceForstall:
Current build failures, building Linux llvm-tblgen:
```
[ 66%] Configuring NATIVE LLVM...
CMake Deprecation Warning at /__w/1/s/src/llvm-project/cmake/Modules/CMakePolicy.cmake:6 (cmake_policy):
  The OLD behavior for policy CMP0116 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
Call Stack (most recent call first):
  CMakeLists.txt:8 (include)


-- The C compiler identification is unknown
-- The CXX compiler identification is unknown
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/cc
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - failed
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc - broken
CMake Error at /usr/share/cmake-3.30/Modules/CMakeTestCCompiler.cmake:67 (message):
  The C compiler

    "/usr/bin/cc"

  is not able to compile a simple test program.
```

The precise compiler is specified for the first cmake invocation and build, but somehow there's a nested cmake invocatoin that is not getting those settings.

AndyAyersMS:
Why are we trying to build tblgen as a separate step?

PR #417, opened 2025-03-26
Az3 coredistools build

AndyAyersMS:


AndyAyersMS:
Hmm, timed out

BruceForstall:
I think the "unified" build of tblgen might work for x64, but not for arm/arm64, since the container we use to build those is a cross-build container without x64-targeting tools, thus, we won't be able to build an x64 tblgen to run. It therefore makes sense to build all the Linux versions the same way, by cross-building tblgen for x64, then cross-building coredistools for x64/arm/arm64. Technically, I think they are all "cross" builds even though the x64 one is building for x64.

BruceForstall:
fwiw, I've tried to learn things about building LLVM from how our Docker cross-build containers are built: https://github.com/dotnet/dotnet-buildtools-prereqs-docker, src\azurelinux\3.0\net10.0\cross\amd64\Dockerfile, src\azurelinux\3.0\net10.0\crossdeps-builder\amd64\Dockerfile

AndyAyersMS:
My previous attempts weren't really cross building, which is why it only worked on x64.

Seems like there is an `llvm-tblgen` next to `clang`, at least on the x64 system, so use that, and change how we specify crossbuilding to the "official" `CMAKE_SYSTEM_NAME=`

AndyAyersMS:
Ah, maybe this will fail too... we'll see.

BruceForstall:
The build completed, including the nuget packaging! I haven't seen that in years.

However, some things worry me. I see:
```
Using C compiler: /usr/local/bin/clang
Using C++ compiler: /usr/local/bin/clang++
Using llvm-tablegen: /usr/local/bin/llvm-tblgen
```
in all the builds. That seems wrong; shouldn't it be pointing at the cross-compiler ROOTFS? e.g., /crossrootfs/x64/...? Or is there some magic where the crossrootfs gets mapped to /usr/local/bin?

Secondly, I think we need to be building our own llvm-tblgen. I presume we need it built from the same LLVM source code we choose to base coredistools on, and not whatever the build toolset docker container has. It turns out right now that both are the same version (20.1.0), but that is not expected to be the normal case. I presume also that the "LLVMTargetsToBuild" could be different (but maybe the built-in one is "all possible targets"?)

BruceForstall:
> The build tools need to be native to the build host, otherwise how can they work properly?

ah, right, I was confused about what is in /crossrootfs -- mostly just headers and libs, not tools

No, we don't need a tblgen that runs on the target.

BruceForstall:
I tested this:

- tested spmi asmdiffs with existing then new coredistools on win/x64. Got same results
- tested spmi asmdiffs with existing then new coredistools on linux/x64 (WSL2, Ubuntu 24.04). Got same results
- Enabled RISC-V use of superpmi, tested asmdiffs with altjit/cross compiler on win/x64
- verified RISC-V coredistools-generated disasm matches (essentially) JitDisasm

build-coredistools.sh
```diff
@@ -47,13 +47,13 @@ fi
...
 echo "Using C compiler: $C_COMPILER"
 echo "Using C++ compiler: $CXX_COMPILER"
+echo "Using llvm-tablegen: $TblGenTool"
 
 case "$TargetOSArchitecture" in
     linux-arm)
         CMakeCrossCompiling=ON
         LLVMDefaultTargetTriple=thumbv7-linux-gnueabihf
         LLVMHostTriple=arm-linux-gnueabihf
-        LLVMTargetsToBuild="ARM"
```
BruceForstall, line 56: Why remove this? The idea here was that linux/arm hosted build probably didn't need the ability to disassemble arm64/x86/x64/LA64/RISCV in our usage. I suppose it doesn't matter if it can, except that make the binary larger, etc.

BruceForstall (approved):

BruceForstall:
I note that the newly built ones are significantly bigger:
```
Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\linux-arm\native
04/04/2025  02:49 PM        15,567,064 libcoredistools.so

Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\linux-arm64\native
04/04/2025  02:49 PM        15,727,840 libcoredistools.so

Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\linux-x64\native
04/04/2025  02:49 PM        15,921,688 libcoredistools.so

Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\osx-arm64\native
04/04/2025  02:49 PM        15,371,592 libcoredistools.dylib

Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\osx-x64\native
04/04/2025  02:49 PM        15,510,424 libcoredistools.dylib

Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\win-arm64\native
04/04/2025  02:49 PM        15,901,184 coredistools.dll

Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\win-x64\native
04/04/2025  02:49 PM        15,775,744 coredistools.dll

Directory of c:\bugs\test_cordistools\Microsoft.NETCore.CoreDisTools.1.4.0\runtimes\win-x86\native
04/04/2025  02:49 PM         9,956,864 coredistools.dll
```
versus (old):
```
 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\linux-arm\native
01/02/2024  09:12 PM         1,734,908 libcoredistools.so

 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\linux-arm64\native
01/02/2024  09:15 PM         8,345,584 libcoredistools.so

 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\linux-x64\native
01/02/2024  09:15 PM         8,500,744 libcoredistools.so

 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\osx-arm64\native
01/02/2024  09:12 PM         8,270,312 libcoredistools.dylib

 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\osx-x64\native
01/02/2024  09:12 PM         8,469,856 libcoredistools.dylib

 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\win-arm64\native
01/02/2024  09:11 PM         8,442,880 coredistools.dll

 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\win-x64\native
01/02/2024  09:11 PM         8,347,136 coredistools.dll

 Directory of C:\.tools\.nuget\packages\microsoft.netcore.coredistools\1.4.0\runtimes\win-x86\native
01/02/2024  09:11 PM         5,863,936 coredistools.dll
```

build-coredistools.sh
AndyAyersMS: That was inadvertent.

AndyAyersMS:
> The addition of LoongArch64 and RISC-V is at least part of this.

Also we gained APX/AVX10 capabilities and maybe more?

PR #422, opened 2025-08-28
Add controlflowguard support to coredistools

agocke:


AndyAyersMS (commented):
Do we need to pass this to the compiler as well?

AndyAyersMS:
```
cl : command line  warning D9025: overriding '/GS' with '/guard:cf>' [D:\a\1\s\obj\win-arm64\tools\coredistools\coredistools.vcxproj]
  coredistools.cpp
D:\a\1\s\src\coredistools\coredistools.cpp(1,1): error C1083: Cannot open include file: 'guardcfw.h': No such file or directory [D:\a\1\s\obj\win-arm64\tools\coredistools\coredistools.vcxproj]
```
Odd. Maybe some target-specific includes are needed?

agocke:
I might do that in the pr because I couldn’t figure out an easy way to install the dependencies to build this locally :)

AndyAyersMS (approved):

PR #424, opened 2025-09-08
Update README to clarify ulimit requirement on macOS

xtqqczze:
This PR updates the `README.md` to clarify that raising the file descriptor limit with:

```sh
ulimit -n 2048
```

is only required on macOS versions prior to 14 (Sonoma).

* Older macOS releases default to a soft limit of 256 file descriptors, which can cause dotnet restore to fail during the bootstrap.sh build.
* Starting with macOS 14 (Sonoma), the default soft limit was increased to 2560, so manually setting the limit is no longer necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ Current tools include:
     cd jitutils
     bootstrap.cmd
 ```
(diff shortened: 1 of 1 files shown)


xtqqczze:
@AndyAyersMS please could you review

xtqqczze:
Support for macOS 13 ended on 15 Sep 2025, so the ulimit requirement note could now be removed completely.

PR #429, opened 2025-09-10
Add ReJIT analysis to instructions retired explorer

AndyAyersMS:
Add new option `-rejit` that dumps info from the tiered compilation events. Also aggregate profiles during each interval delinated by these events to show how the runtime cpu samples distribute across managed code from different tiers.

Sample output (charted) showing how Tier1 code (green dots) eventually becomes prominent as rejitting progresses (orange line), but that we also never drain the rejit queue (blue line):

image

You can run this on ETL profiles gathered by perfview or the `-p ETL` option to BenchmarkDotNet.

Rejitting interval and benchmark interval analysis are not yet aware of each other, though it would be a worthy project (in particualr seeing if the engine actual intervals have any appreciable fraction of non-tier1 code).


AndyAyersMS:
FYI @dotnet/jit-contrib

PR #432, opened 2025-09-19
Add updated R2R analysis

AndyAyersMS:
We no longer get method load events for R2R methods, but there are events that are fired when the runtime goes to find an R2R entry point that are almost as good. Enable these via `-RuntimeLoading` command line arg to perfview.

Add support for parsing these events and inferring the length of R2R methods from them.

### DIFF
--- a/src/instructions-retired-explorer/instructions-retired-explorer.cs
+++ b/src/instructions-retired-explorer/instructions-retired-explorer.cs
@@ -489,21 +489,24 @@ namespace CoreClrInstRetired
             List<ImageInfo> r2rAssembyImages = new List<ImageInfo>();
             foreach (ImageInfo i in ImageMap.Values)
             {
+                // When parsing ETL, we used size zero for methods whose size were not known.
+                // Collect these up so we can determine an approximate size (good enough for sample attribution).
+                //
                 if (i.IsJitGeneratedCode && i.Tier == OptimizationTier.ReadyToRun && i.Size == 1)
                 {
                     r2rMethodImages.Add(i);
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


AndyAyersMS:
@dotnet/jit-contrib PTAL

src/instructions-retired-explorer/instructions-retired-explorer.cs
```diff
@@ -476,6 +483,75 @@ static void GroupMethods()
             }
         }
 
+        static void FixR2RLengths()
+        {
+            List<ImageInfo> r2rMethodImages = new List<ImageInfo>();
+            List<ImageInfo> r2rAssembyImages = new List<ImageInfo>();
+            foreach (ImageInfo i in ImageMap.Values)
+            {
+                if (i.IsJitGeneratedCode && i.Tier == OptimizationTier.ReadyToRun && i.Size == 1)
```
kg, line 492: Should we have a comment explaining why `Size == 1`? Seems magical

src/instructions-retired-explorer/instructions-retired-explorer.cs
```diff
@@ -476,6 +483,75 @@ static void GroupMethods()
...
+                {
+                    r2rMethodImages.Add(i);
+                }
+                else
+                {
+                    r2rAssembyImages.Add(i);
+                }
+            }
+
+            ImageInfo[] methodArray = new ImageInfo[r2rMethodImages.Count];
```
kg, line 502: nit: Can we just use List.ToArray for this? Or is there a specific reason why it needs to be this way

kg:
LGTM

JulieLeeMSFT (approved):

PR #438, opened 2026-05-29
Adding Antigen source to jitutils

dhartglassMSFT:
- Imported code using `git subtree add ...` without a squash
- After this goes in we'll change fuzzer_setup.py (runtime repo) to instead point here
- git log graph looks like the following (69d25... and 8fb27a etc are antigen commits, 942ef1... is jitutils)

```
* 99674a2 Update jitutils README to point to Antigen
*   8b47f27 Add 'src/Antigen/' from commit '69d2547f91a233e6002001e5ca149e8661a4716a'
|\  
| * 69d2547 fix build failure
| * 8fb27ab fix the CompileError for SVE
...
| * 2e497f1 Initial commit
* 942ef19 Change jitdiff parallelism from coreCount * 2 to coreCount (#436)
```


AndyAyersMS (approved):
Just a cursory review, since we want to import this as is.

PR #439, opened 2026-06-15
pmi: run all class constructors up front for deterministic -cctors output

EgorBo:
## Problem

PMI-based jit-diff (e.g. MihuBot's `-nuget` runs) produces noisy diffs even on PRs that change nothing — see https://github.com/MihuBot/runtime-utils/issues/1981. The diffs show types that are *randomly* already statically initialized (or not) by the time a method is compiled, which changes inlining, class-init helpers and `static readonly` folding between the base and diff runs.

## Root cause

The JIT specializes codegen based on whether a referenced type is already initialized (`initClass` / `getStaticFieldContent`): once a type's `.cctor` has run, the JIT can drop class-init helpers, fold `static readonly` fields and inline more aggressively. `PrepareMethod` itself does **not** run any `.cctor`, so the JIT observes whatever init state happened to exist incidentally at that moment (runtime startup, PMI's own reflection, background-thread activity). Under the parallel load of a jit-diff run that timing shifts, so the same method sees a type as initialized in one run and not in another — i.e. non-deterministic output.

## Fix

When PMI runs with `-cctors` (which MihuBot passes by default), initialize **every loaded type up front, before any method is JIT'd**, instead of lazily/per-type. This walks the target assembly's whole reference closure — most importantly `System.Private.CoreLib` — and runs every concrete type's `.cctor`. With all classes deterministically initialized, `initClass` consistently reports INITIALIZED and codegen no longer depends on init timing.

The existing per-type cctor path is kept because it also initializes the **closed generic instantiations** PMI creates during the walk (e.g. `Dictionary<string,int>`), which the up-front pass can't enumerate. The two are complementary.

Notes:
- Open generic definitions are skipped; `.cctor` failures are swallowed so initialization is exhaustive and order-independent.
- Only affects `-cctors` mode; default behavior is unchanged.
- The extra CoreLib methods jitted by running cctors are filtered out of the dasm by `DOTNET_JitDisasmAssemblies` in jit-dasm-pmi, so they don't appear in diffs.

cc @MihaZupan

### DIFF
--- a/src/pmi/pmi.cs
+++ b/src/pmi/pmi.cs
@@ -938,6 +938,108 @@ class Worker
         return maxResult;
     }
 
+    // When running with "cctors first", initialize every loaded type up front - before any method
+    // is JIT'd - rather than lazily / per-type. This covers the target assembly and every referenced
+    // assembly, most importantly System.Private.CoreLib. The JIT specializes codegen based on whether
+    // a referenced type is already initialized (initClass / getStaticFieldContent): when a type's
+    // .cctor has run it can drop class-init helpers, fold 'static readonly' fields and inline more
+    // aggressively. Whether some incidental or background-thread execution happened to run a given
+    // .cctor before a particular method was compiled is timing-dependent, which is the main source of
+    // non-deterministic PMI diffs under the parallel load of a jit-diff run. Forcing a fully and
+    // deterministically initialized class state removes that variability.
+    private void RunAllClassConstructorsFirst(Assembly target)
+    {
+        // Pull in the whole reference closure first so that types referenced only from method bodies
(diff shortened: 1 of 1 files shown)


MihaZupan (approved):

EgorBo:
@dotnet/jit-contrib can someone take a look? this fixes non-determinism for `@MihuBot -nuget` (where it runs diffs on a big set of OSS libs).

Basically, instead of "visit a class, run its cctor, jit its methods" we have a pre-pass where we unconditionally initialize all classes first (in `-cctors` mode only). This kills very verbose/non-deterministic diffs

Before this PR: https://github.com/MihuBot/runtime-utils/issues/1984
After this PR: https://github.com/MihuBot/runtime-utils/issues/1985

PR #440, opened 2026-09-02
Allow parsing code size without prolog metrics

EgorBo:
Decouple `CodeSize` parsing from `PrologSize` so jit-analyze accepts a minimal JIT metrics line containing only code size and method identity.

Checked JIT output continues to provide optional prolog and allocation metrics. When allocation size is absent, jit-analyze now treats the allocation metrics as unavailable instead of producing a negative `ExtraAllocBytes` value or triggering a debug assertion.

Validation:
- `dotnet build -c Release src/jit-analyze/jit-analyze.csproj`
- Parsed minimal Release and rich Checked JIT footer formats

### DIFF
--- a/src/jit-analyze/Program.cs
+++ b/src/jit-analyze/Program.cs
@@ -292,7 +292,8 @@ namespace ManagedCodeGen
         public static IEnumerable<MethodInfo> ExtractMethodInfo(string[] filePaths)
         {
             Regex namePattern = new Regex(@"for method (.*)$");
-            Regex codeAndPrologSizePattern = new Regex(@"code ([0-9]{1,}), prolog size ([0-9]{1,})");
+            Regex codeSizePattern = new Regex(@"^; Total bytes of code ([0-9]{1,}).* for method ");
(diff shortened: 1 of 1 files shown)


EgorBo:
@dotnet/jit-contrib @MihaZupan can someone please review?

It doesn't fix jit-utils to use Release JIT yet, but doesn't force the regex to match unnecessary information.
I want to unify JitDisasm in Checked and Release in dotnet/runtime

Checked:
```
; Total bytes of code 42, prolog size 5, PerfScore 10.25, instruction count 12, allocated bytes for code 48 (MethodHash=12345678) for method Foo:Bar() (FullOpts)
```
Release:
```
; Total bytes of code 42
```
I think we need to add method name to footer for Release for simpler parsing:
```
; Total bytes of code 42 for method Foo:Bar() (FullOpts)
```

https://github.com/dotnet/runtime/pull/133095

I could also fix the parser only in jit-diffs (not a trivial fix), but I think having method name in footer might be handy

MihaZupan (approved):

EgorBo:
Merging since this is not a change in behavior

